PR #7, opened 2020-08-03
Added datastore functionality to Classroom, User, Form, and DatabaseService

kyleasmith1:
Added methods inside of Classroom that would allow teachers to add/remove students and add/remove forms, stored everything inside of the classes as datastore keys, and created a DatabaseService that allows you to create objects with a given key and save entities to datastore.


capstone/src/main/java/com/google/sps/data/Classroom.java
```diff
@@ -1,94 +1,100 @@
...
-    public void setTeacher(Teacher teacher) {
-        this.teacher = teacher;
+    public Classroom(Entity entity) {
+        this.entity = entity;
     }
 
-    public void setSubject(String subject) {
-        this.subject = subject;
+    public Classroom(User teacher, String subject) { 
+        this.entity = new Entity(CLASSROOM);
```
mattlgy: Just calling it out, no change needed. Saying `CLASSROOM` and `Classroom.CLASSROOM' while in the context of the `Classroom` class are synonymous.

capstone/src/main/java/com/google/sps/data/Classroom.java
```diff
@@ -1,94 +1,100 @@
...
-    public Teacher getTeacher() {
-        return this.teacher;
+    
+    public Key getTeacher() {
+        return (Key) this.entity.getProperty(TEACHER);
     }
 
     public String getSubject() {
-        return this.subject;
+        return (String) this.entity.getProperty(SUBJECT);
```
mattlgy: I'm not sure about the `Entity` API, but you _may_ need to check for the null case?

capstone/src/main/java/com/google/sps/data/User.java
```diff
@@ -1,12 +1,43 @@
...
-    public int getId();
-}
+
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.Entity;
+import com.google.sps.service.DatabaseService;
+
+public class User {
+    protected static final String NICKNAME = "nickname";
+    protected static final String USER = "User";
```
mattlgy: This looks like the only Uppercase string-key, is that on purpose?

mattlgy (approved):

capstone/src/main/webapp/config.js
```diff
@@ -0,0 +1,5 @@
+var config = {
```
mattlgy: Should this be checked in?

capstone/src/main/java/com/google/sps/data/Classroom.java
kyleasmith1: got it

capstone/src/main/java/com/google/sps/data/Classroom.java
kyleasmith1: are you saying in the off chance that someone tries to create an object and inputs null values as the parameters?

capstone/src/main/java/com/google/sps/data/User.java
kyleasmith1: Yes, this is kind of entity so I made it uppercase

capstone/src/main/webapp/config.js
kyleasmith1: no, this was fixed (will be fixed) in a later PR... our gitignores messed up in a merge.

capstone/src/main/java/com/google/sps/data/User.java
707090: Having a naming convention on the constants makes that more clear. So something like NICKNAME_PROPERTY_KEY and USER_ENTITY_NAME or whatever else would help readability

capstone/src/main/java/com/google/sps/data/Classroom.java
707090: I will request a change though. When using static variables, you should always preface it with the classname to disambiguate it from instance variables.

capstone/src/main/java/com/google/sps/data/Classroom.java
707090: As I had said in another standup: never assume anything about the chances of things. Your code should operate on as few assumptions as possible, including not assuming the data in the database is good data produced in the currently expected way.

That said, I dont think the null case needs to be checked since (horray... Java 😒 ) null is a valid value for all objects.

capstone/src/main/java/com/google/sps/data/Classroom.java
```diff
@@ -1,94 +1,100 @@
...
+    
+    @SuppressWarnings("unchecked")
+    public void addStudent(User student) {
+        List<Key> students = (ArrayList<Key>) this.entity.getProperty(STUDENTS);
+        students.add(student.getUserKey());
 
-    // Database
-    public void addStudent(Student student) { 
-        students.add(student);
+        this.entity.setProperty(STUDENTS, students);
```
707090: Have you tried the code without this line? Im not sure which type the API exposes, but it might A: be copying in which case this is necessary, or B: return the backing list or a list subclass which affects the underlying entity when you manipulate the list

capstone/src/main/java/com/google/sps/data/Classroom.java
```diff
@@ -1,94 +1,100 @@
...
+
+        this.entity.setProperty(FORMS, forms);
     }
 
-    public boolean isStudentInClass(Student student) {
+    @SuppressWarnings("unchecked")
+    public boolean isStudentInClass(User student) {
+        List<Key> students = (ArrayList<Key>) this.entity.getProperty(STUDENTS);
+
         assert students.size() > 0: "There are no students in this class!";
```
707090: Asserts should be used when you think that the case only happens when a programmer has made an error. So if you have something like:
// Hey developers, you should only call this with positive numbers
public double sqrt(double input) {}

Then that would be an scenario where you would assert that the input is positive. In this case, classrooms legitimately can have no students and thats normal, it just means that you always return false. (But you dont actually need a separate check for that, since the rest of your logic applies equally well to an empty list

capstone/src/main/java/com/google/sps/data/Classroom.java
```diff
@@ -1,94 +1,100 @@
...
 
-        if (students.lastIndexOf(student) == -1) {
+        if (students.lastIndexOf(student.getUserKey()) == -1) {
             return false;
         } else {
             return true;
         }
     }
```
707090: Whenever you see the pattern:
if (somecondition) {
    return true;
} else {
   return false;
}
you can simplify to 
return somecondition;
(or in your case !somecondition)

capstone/src/main/java/com/google/sps/data/RequestParser.java
```diff
@@ -22,4 +22,18 @@
...
+        try {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                sb.append(line).append('\n');
+            }
+        } finally {
+            reader.close();
+        }
+        return sb.toString();
+    }
```
707090: Keeping these separate creates code duplication. Instead, the more complicated version of the method (parseObject) should delegate to the simpler version:
public static <T> T parseObjectFromRequest(HttpServletRequest request, Class<T> type) throws IOException {
        Gson gson = new Gson();
        T object = gson.fromJson(parseStringFromRequest(request), type);
        return object;
    }

Also see my comment on the original PR #5 about this method

capstone/src/main/webapp/form-loader.js
```diff
@@ -48,21 +48,21 @@ function getForms() {
...
+function createTeacherFormElement(editUrl) { 
     const aElement = document.createElement("a");
     aElement.innerText = "Form Edit Page";
-    aElement.href = editURL;
+    aElement.href = editUrl;
     return aElement;
 }
 
-function createStudentFormElement(URL) {
+function createStudentFormElement(Url) {
```
707090: for variable naming, you still start with lowercase even if it is an acronym. (you got the editUrl one right though)
(applies everywhere)

capstone/src/main/java/com/google/sps/servlets/FormHandlerServlet.java
```diff
@@ -5,28 +5,32 @@
...
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        List<Form> forms = new ArrayList<>();
-        Form form = null;
         Query query = new Query ("Form");
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery results = datastore.prepare(query);
+        
+        List<Form> forms = new ArrayList<>();
+        Form form = null;
```
707090: Inline

There is no need to declare a forward variable if it is only used once in the following line. So you can simplify your adding to just say 
forms.add(new Form(entity));

707090 (changes requested):

capstone/src/main/java/com/google/sps/data/Classroom.java
kyleasmith1: sounds good

capstone/src/main/java/com/google/sps/data/Classroom.java
kyleasmith1: we can talk about this more in the standup about how to handle null inputs

capstone/src/main/java/com/google/sps/data/Classroom.java
kyleasmith1: got it

capstone/src/main/java/com/google/sps/data/Classroom.java
kyleasmith1: resolved: will add in a future PR

capstone/src/main/java/com/google/sps/data/RequestParser.java
```diff
@@ -8,18 +8,8 @@
...
-            String line;
-            while ((line = reader.readLine()) != null) {
-                sb.append(line).append('\n');
-            }
-        } finally {
-            reader.close();
-        }
         Gson gson = new Gson();
-        T object = gson.fromJson(sb.toString(), type);
+        T object = gson.fromJson(parseStringFromRequest(request), type);
```
707090: Missed an inline

capstone/src/main/java/com/google/sps/servlets/FormHandlerServlet.java
```diff
@@ -25,20 +25,16 @@
...
         for(Entity entity : results.asIterable()){
-            form = new Form(entity);
-            forms.add(form);
+            forms.add(new Form(entity));
         }
     
-        Gson gson = new Gson();
-        String json = gson.toJson(forms);
+        String json = (new Gson()).toJson(forms);
         response.setContentType("application/json");
```
707090: Same as above

707090 (approved):

capstone/src/main/java/com/google/sps/data/RequestParser.java
AndrewDPham: Got it

capstone/src/main/java/com/google/sps/servlets/FormHandlerServlet.java
AndrewDPham: Got it.

PR #8, opened 2020-08-07
Create login functionality

kyleasmith1:
Users can now create accounts on the web application (have User objects stored in datastore specific to them). 

Other Things:
Inline fixes
Removed unnecessary imports
Got rid of duplicate lines of code
Added error responses

### DIFF
--- a/capstone/src/main/java/com/google/sps/data/Form.java
+++ b/capstone/src/main/java/com/google/sps/data/Form.java
@@ -17,30 +17,30 @@ public class Form {
     }
 
     public Form(String editUrl, String Url) {
-        this.entity = new Entity(FORM);
-        this.entity.setProperty(EDIT_URL, editUrl);
-        this.entity.setProperty(URL, Url);
-        this.entity.setProperty(PUBLISHED, false);
+        this.entity = new Entity(Form.FORM);
+        this.entity.setProperty(Form.EDIT_URL, editUrl);
+        this.entity.setProperty(Form.URL, Url);
+        this.entity.setProperty(Form.PUBLISHED, false);
     }
 
     public void publish() {
-        this.entity.setProperty(PUBLISHED, true);
+        this.entity.setProperty(Form.PUBLISHED, true);
     }
 
     public void unPublish() {
-        this.entity.setProperty(PUBLISHED, false);
+        this.entity.setProperty(Form.PUBLISHED, false);
     }
 
     public String getEditURL() {
-        return (String) this.entity.getProperty(EDIT_URL);
+        return (String) this.entity.getProperty(Form.EDIT_URL);
     }
 
     public String getURL() {
-        return (String) this.entity.getProperty(URL);
+        return (String) this.entity.getProperty(Form.URL);
     }
 
     public boolean publishState() {
-        return (Boolean) this.entity.getProperty(PUBLISHED);
+        return (Boolean) this.entity.getProperty(Form.PUBLISHED);
     }
 
     public Entity getFormEntity() {
--- a/capstone/src/main/java/com/google/sps/data/Form.java
+++ b/capstone/src/main/java/com/google/sps/data/Form.java
@@ -6,7 +6,7 @@ import com.google.sps.service.DatabaseService;
 
 public class Form {
     protected static final String FORM = "Form";
-    protected static final String EDIT_URL = "editUrl";
+    protected static final String EDIT = "editUrl";
     protected static final String URL = "Url";
     protected static final String PUBLISHED = "published";
 
@@ -18,7 +18,7 @@ public class Form {
 
     public Form(String editUrl, String Url) {
         this.entity = new Entity(Form.FORM);
-        this.entity.setProperty(Form.EDIT_URL, editUrl);
+        this.entity.setProperty(Form.EDIT, editUrl);
         this.entity.setProperty(Form.URL, Url);
         this.entity.setProperty(Form.PUBLISHED, false);
     }
@@ -32,7 +32,7 @@ public class Form {
     }
 
     public String getEditURL() {
-        return (String) this.entity.getProperty(Form.EDIT_URL);
+        return (String) this.entity.getProperty(Form.EDIT);
     }
 
     public String getURL() {
--- a/capstone/src/main/java/com/google/sps/data/Form.java
+++ b/capstone/src/main/java/com/google/sps/data/Form.java
@@ -5,42 +5,42 @@ import com.google.appengine.api.datastore.Entity;
 import com.google.sps.service.DatabaseService;
 
 public class Form {
-    protected static final String FORM = "Form";
-    protected static final String EDIT = "editUrl";
-    protected static final String URL = "Url";
-    protected static final String PUBLISHED = "published";
+    public static final String FORM_ENTITY_NAME = "Form";
+    public static final String EDIT_URL_PROPERTY_KEY = "editUrl";
+    public static final String URL_PROPERTY_KEY = "url";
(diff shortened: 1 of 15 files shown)


capstone/src/main/webapp/script.js
```diff
@@ -15,9 +15,13 @@
 function onSignIn(googleUser) {
   var profile = googleUser.getBasicProfile();
   var id_token = googleUser.getAuthResponse().id_token;
-  console.log(profile.getEmail() + " has signed in!");
-
-  window.location.href = "form.html";
+  return fetch("/login", {method: "POST", body: id_token}).then((resp) => {
+        if (resp.ok){
```
kyleasmith1: We're going to fix this in our bad data PR. Having this "if (resp.ok)"  can cause errors because there's a chance that "resp" is undefined.

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,68 @@
...
+                String name = (String) idToken.getPayload().get(LoginServlet.NAME_PROPERTY_KEY);
+
+                DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+                Filter userFilter = new FilterPredicate(User.USER_ID_PROPERTY_KEY, FilterOperator.EQUAL, userId);
+                Query query = new Query(User.USER_ENTITY_NAME).setFilter(userFilter);
+                PreparedQuery results = datastore.prepare(query);
+
+                if (results.countEntities() == 0) {
+                    DatabaseService.save(new User(userId, name).getUserEntity());
```
kyleasmith1: Andrew and I were talking about maybe adding a "savableEntity" method that can extract the entity instead of having to call "object.get___Entity()" anytime we want to save something

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,68 @@
...
+
+@WebServlet("/login")
+public class LoginServlet extends HttpServlet {
+    private static final String NAME_PROPERTY_KEY = "name";
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {    
+        try {
+            GoogleIdToken idToken = tokenVerifier().verify(RequestParser.parseStringFromRequest(request));
+            if (idToken != null) {
```
mattlgy: To avoid nesting further to the right, I'd recommend writing this as:

```java
if (idToken == null) {
    response.setContentType("text/html");
    response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
    return;
}

String userId = idToken.getPayload().getSubject();
...
```

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/login")
+public class LoginServlet extends HttpServlet {
+    private static final String NAME_PROPERTY_KEY = "name";
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {    
+        try {
```
mattlgy: This is a lot of code to nest under a single try block. Would it be possible to just have the try ... catch around the line(s) that can actually throw an error?

capstone/src/main/webapp/script.js
mattlgy: `if (resp?.ok)` Would be a bit safer and not too much of a change for this PR.

mattlgy (changes requested):

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
kyleasmith1: I actually need to keep everything in try block because Java doesn't let me reference things outside of the try block.

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
kyleasmith1: Got it

capstone/src/main/webapp/script.js
kyleasmith1: Got it

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,72 @@
...
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {  
+        GoogleIdToken idToken = null;
+
+        try {
+           idToken = tokenVerifier().verify(RequestParser.parseStringFromRequest(request));
+        } catch (GeneralSecurityException e){
+            System.out.println("Cannot verify token: " + e);
+        }
+
+        if (idToken == null) {
```
mattlgy: I _think_ you can do this in the catch statement.

mattlgy (approved):

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
kyleasmith1: got it

capstone/src/main/java/com/google/sps/data/Form.java
```diff
@@ -15,7 +15,7 @@
     public Form(Entity entity) {
         this.entity = entity;
     }
-
+
```
707090, line 18: Unnecessary whitespace. Also if you do want to make little formatting changes like this, its easier to lump them all up into one big formmatting CL full of meaningless changes so its easier to review. (because sadly github cant tell this is an irrelevant change). That helps it so that I can just whiz through a formatting CL in 2 seconds and it doesnt clutter CLs with actual content

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {  
+        GoogleIdToken idToken = null;
+
+        try {
+           idToken = tokenVerifier().verify(RequestParser.parseStringFromRequest(request));
+        } catch (GeneralSecurityException e){
+            System.out.println("Cannot verify token: " + e);
+            if (idToken == null) {
+                response.setContentType("text/html");
```
707090: You never need to set the content type if that is not what you are returning (applies everywhere)

capstone/src/main/webapp/index.html
```diff
@@ -2,16 +2,15 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <title>Home</title>
+    <title>Login</title>
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
     <script src="https://apis.google.com/js/platform.js" async defer></script>
-    <meta name="google-signin-client_id" content="">
+    <meta name="google-signin-client_id" content=''>
```
707090, line 9: This is in a similar vein of no-op changes that can be lumped together

707090 (changes requested):

capstone/src/main/java/com/google/sps/data/Form.java
kyleasmith1: Yea this is my bad. I'll keep the meaningless changes CL thing in mind for future CLs.

capstone/src/main/java/com/google/sps/servlets/LoginServlet.java
kyleasmith1: Got it

PR #9, opened 2020-08-07
Allow users to create classrooms and store on database

AndrewDPham:
Simple postHandler to create classrooms and store on database. For now, we are creating dummy users and sending user email's as userIDs.

### DIFF
--- a/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
@@ -26,26 +26,6 @@ import com.google.sps.service.DatabaseService;
 @WebServlet("/dashboard-handler")
 public class DashboardHandlerServlet extends HttpServlet {
 
-    // @Override
-    // public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    //     List<Form> forms = new ArrayList<>();
-    //     Form form = null;
-    //     Query query = new Query ("Form");
-    //     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    //     PreparedQuery results = datastore.prepare(query);
-    
-    //     for(Entity entity : results.asIterable()){
-    //         form = new Form(entity);
-    //         forms.add(form);
-    //     }
-    
-    //     Gson gson = new Gson();
-    //     String json = gson.toJson(forms);
-    //     response.setContentType("application/json");
-    //     response.getWriter().println(json);
-    // }
-
-
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
@@ -64,12 +44,12 @@ public class DashboardHandlerServlet extends HttpServlet {
         JsonElement jelement = JsonParser.parseString(sb.toString());
         JsonObject jobject = jelement.getAsJsonObject();
 
-        User testUser = new User("Example.com", "John");
+        String name = jobject.get("name").getAsString();
+        String email = jobject.get("email").getAsString();  
+        User testUser = new User(email, name);
         DatabaseService.save(testUser.getUserEntity());
 
         String subject = jobject.get("subject").getAsString();
-        System.out.println(subject);
-        
         Classroom classroom = new Classroom(testUser, subject);
         DatabaseService.save(classroom.getClassroomEntity());
 
--- a/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
@@ -14,7 +14,7 @@ import com.google.gson.JsonParser;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.sps.data.Form;
-import com.google.sps.data.RequestJsonParser;
+import com.google.sps.data.RequestParser;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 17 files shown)


capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -0,0 +1,21 @@
+document.body.prepend(dynamicButton);
+
+function createClassroom() {
+    var auth2 = gapi.auth2.getAuthInstance();
+    var name = "";
+    var email = "";
+    if (auth2.isSignedIn.get()) {
+        var name = auth2.currentUser.get().getBasicProfile().getName();
```
kyleasmith1: Do we need a name variable for classrooms?

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -0,0 +1,21 @@
...
+        var email = auth2.currentUser.get().getBasicProfile().getEmail();
+    }
+
+    var subject = document.querySelector('#subject').value;
+    var classroomData = JSON.stringify({ "nickname": name, "userId": email, "subject": subject });
```
kyleasmith1, line 12: "userId: email" seems kind of confusing

AndrewDPham: Yeah, for now I'm using their email as their ID since I don't to pass their user ID to the backend.

capstone/src/main/webapp/dashboard-loader.js
AndrewDPham: This is to create the dummy user for now. The user dummy get its name here.

capstone/src/main/webapp/dashboard-loader.js
kyleasmith1, line 12: Inline

AndrewDPham: Got it. I think I'll leave it for now since inline would create a line with 100 characters.

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
```diff
@@ -0,0 +1,33 @@
...
+public class DashboardHandlerServlet extends HttpServlet {
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+        JsonObject jobject = JsonParser.parseString(RequestParser.parseStringFromRequest(request)).getAsJsonObject();
+
+        User teacher = new User(jobject.get(User.USER_ID_PROPERTY_KEY).getAsString(),
+            jobject.get(User.NICKNAME_PROPERTY_KEY).getAsString());
+        DatabaseService.save(teacher.getUserEntity());
```
mattlgy, line 26: This seems like it will create a new user every time, even if the logged in person already has a user object saved?

AndrewDPham: Yeah, this a dummy user for now. We will keep track of real users once we have the web filter! :)

707090: Making shortcuts like these make the PRs more digestible and easy to organize, but its worth adding a comment above the code just to clarify

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
```diff
@@ -0,0 +1,33 @@
...
+
+        Classroom classroom = new Classroom(teacher, jobject.get(Classroom.SUBJECT_PROPERTY_KEY).getAsString());
+        DatabaseService.save(classroom.getClassroomEntity());
+
+        response.setContentType("text/html");
```
707090: If you are not returning HTML, there is no need to set this content type

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -0,0 +1,21 @@
...
+function createClassroom() {
+    var auth2 = gapi.auth2.getAuthInstance();
+    var name = "";
+    var email = "";
+    if (auth2.isSignedIn.get()) {
+        name = auth2.currentUser.get().getBasicProfile().getName();
+        email = auth2.currentUser.get().getBasicProfile().getEmail();
+    }
+
+    var subject = document.querySelector('#subject').value;
```
707090: Having a reference to a particular DOM structure right in the method reduces the reusability of this function to just this page. Instead you could pass the subject as a parameter to the function, and this function suddenly becomes re-usable by potentially any part of the app.

707090: When you are running document.X, it is more clear to use the getElementById function than the general purpose querySelector (and more efficient since that has to run the CSS parser and may bring back potentially any number of nodes)

707090 (changes requested):

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
AndrewDPham: Sounds good. I added it in the description of the PR but it's probably better to add the comments here.

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
AndrewDPham: Sounds good.

capstone/src/main/webapp/dashboard-loader.js
AndrewDPham: Hmm sounds good!

AndrewDPham: Sounds good

kyleasmith1: got it

kyleasmith1: got it

707090 (approved):

PR #10, opened 2020-08-07
Display and Get Classroom data

AndrewDPham:
Display and get classroom data. Will change createClassroomDivElement() in a later branch to display with Bootstrap styling. For now, it displays IDs of the dummy User and Classroom.

### DIFF
--- a/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
@@ -21,7 +21,7 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-@WebServlet("/dashboard-handler")
+@WebServlet("/dashboard")
 public class DashboardHandlerServlet extends HttpServlet {
 
     @Override
--- a/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
@@ -27,6 +27,25 @@ import com.google.sps.service.DatabaseService;
 public class DashboardHandlerServlet extends HttpServlet {
 
     @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        List<Classroom> classrooms = new ArrayList<>();
+        Classroom classroom = null;
+        Query query = new Query("Classroom");
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery results = datastore.prepare(query);
+    
+        for(Entity entity : results.asIterable()){
+            classroom = new Classroom(entity);
+            classrooms.add(classroom);
+        }
+    
+        Gson gson = new Gson();
+        String json = gson.toJson(classrooms);
+        response.setContentType("application/json");
+        response.getWriter().println(json);
+    }
+
+    @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
         BufferedReader reader = request.getReader();
--- a/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
@@ -30,8 +30,10 @@ public class DashboardHandlerServlet extends HttpServlet {
         
         // TODO: In a later CL we will make it so that it only gets User-specific classrooms
         PreparedQuery results = datastore.prepare(new Query(Classroom.CLASSROOM_ENTITY_NAME));
+
+       // List<Entity> classrooms = results.asList();
     
-        List<Classroom> classrooms = new ArrayList<>();
+        ArrayList<Classroom> classrooms = new ArrayList<>();
         for(Entity entity : results.asIterable()){
             classrooms.add(new Classroom(entity));
         }
--- a/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
@@ -26,21 +26,16 @@ public class DashboardHandlerServlet extends HttpServlet {
 
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        List<Classroom> classrooms = new ArrayList<>();
-        Classroom classroom = null;
-        Query query = new Query("Classroom");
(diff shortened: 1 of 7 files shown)


capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -13,9 +13,41 @@ function createClassroom() {
...
         }
     });
 }
+
+function getClassrooms() {
+    fetch("/dashboard-handler").then(response => response.json()).then((classroomsList) => {
+        const classroomElement = document.getElementById("classroom-container");
+        classroomElement.innerHTML = "";
+        for (classroom of classroomsList) {
+            console.log(classroom);
```
kyleasmith1: take this out

AndrewDPham: Got it.

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -13,9 +13,40 @@ function createClassroom() {
...
+    fetch("/dashboard-handler").then(response => response.json()).then((classroomsList) => {
+        const classroomElement = document.getElementById("classroom-container");
+        classroomElement.innerHTML = "";
+        for (classroom of classroomsList) {
+            classroomElement.appendChild(createClassroomDivElement(classroom));
+        };
+    });
+}
+
+function createClassroomDivElement(classroom) {
```
mattlgy, line 32: Recommend using string templates and either DOMParser or innerHTML.

https://developer.mozilla.org/en-US/docs/Web/API/DOMParser

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
```diff
@@ -15,6 +24,20 @@
 @WebServlet("/dashboard-handler")
 public class DashboardHandlerServlet extends HttpServlet {
 
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery results = datastore.prepare(new Query(Classroom.CLASSROOM_ENTITY_NAME));
```
mattlgy, line 32: This will get every classroom, even if the user is not in them? (Fine for now but I recommend leaving a TODO or note saying as such.)

mattlgy (changes requested):

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
```diff
@@ -15,6 +24,20 @@
...
+    
+        List<Classroom> classrooms = new ArrayList<>();
```
707090: There is no need to upcast this. In this case it doesnt make a difference so there is no need to change it if you dont want, but as a rule, I wouldnt upcast unless there is a reason to

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
```diff
@@ -15,6 +24,20 @@
...
+        for(Entity entity : results.asIterable()){
+            classrooms.add(new Classroom(entity));
+        }
```
707090: You can skip/inline all of this with PreparedQuery.asList https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PreparedQuery.html#asList-com.google.appengine.api.datastore.FetchOptions- (You can create default fetch options with the FetchOptionsBuilder https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/FetchOptions.Builder.html#withDefaults--)

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
```diff
@@ -15,6 +24,20 @@
 @WebServlet("/dashboard-handler")
```
707090: Servlets dont need "-handler" on them since, by definition unless the request gives like 404 or something, you are automatically asking a handler to handle your request.

capstone/src/main/webapp/dashboard-loader.js
707090: Alternatively, this would be a place handlebars.js could be useful

capstone/src/main/webapp/dashboard.html
```diff
@@ -6,14 +6,16 @@
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
-  <body>
+  <body onload="getClassrooms()">
 
     <div id="content">
```
707090, line 11: The IDs here are a little confusing. This implies this is the content of the page, but it looks like its... The creation modal?

707090 (changes requested):

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
AndrewDPham: Yeah, until we get the web filter, we will query all of the classrooms for now.

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
AndrewDPham: Got it.

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
AndrewDPham: Got it.

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
kyleasmith1: TODO comment added

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
kyleasmith1: Removed

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
kyleasmith1: I believe the difference would be that asList("Insert fetch option builer here") would return the classrooms as entities but at least for this implementation we needed the classroom objects (didn't know if you were aware of this/noticed) for the client side of the code.

capstone/src/main/webapp/dashboard.html
kyleasmith1: removed unnecessary div

707090 (approved):

mattlgy (approved):

PR #11, opened 2020-08-12
Filter

AndrewDPham:
Create webFilter while moving helper verify functions for general use. Add web.xml and for now, headers are id_token.

### DIFF
--- a/capstone/src/main/java/com/google/sps/filter/AuthenticateFilter.java
+++ b/capstone/src/main/java/com/google/sps/filter/AuthenticateFilter.java
@@ -30,6 +30,8 @@ import com.google.sps.data.RequestParser;
 @WebFilter("/auth-filter")
 public class AuthenticateFilter implements javax.servlet.Filter {
 
+    private static final String AUTH_HEADER_NAME = "id_token";
+
     public void init(FilterConfig filterConfig) throws ServletException{
 
     }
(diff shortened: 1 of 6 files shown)


capstone/src/main/java/com/google/sps/filter/AuthenticateFilter.java
```diff
@@ -0,0 +1,70 @@
...
+            return;
+        } catch (Exception e){
+            res.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+            return;
+        }
+
+        String userId = idToken.getPayload().getSubject();
+
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Filter userFilter = new FilterPredicate(User.USER_ID_PROPERTY_KEY, FilterOperator.EQUAL, userId);
```
kyleasmith1, line 58: You also use this code in LoginServlet.java. Maybe you could make a small helper function so that you don't have to repeat code. It could be a general use function that takes in two parameters (a property key and an id of some sort) and finds the specific entity with the FilterPredicate.

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -1,17 +1,18 @@
...
 
-function createClassroom() {
+function createClassroom(subject) {
     var auth2 = gapi.auth2.getAuthInstance();
     var name = "";
     var email = "";
+    var id_token = "";
     if (auth2.isSignedIn.get()) {
         name = auth2.currentUser.get().getBasicProfile().getName();
         email = auth2.currentUser.get().getBasicProfile().getEmail();
```
kyleasmith1, line 10: Do we still need a user's email now that we have the web filter?

AndrewDPham: Although the web filter was built, the next PR focuses on creating the client Header functionality so that the PR after that focuses on using the ID_TOKEN to create classrooms that are related to the user. For now, dummy users are still being used.

capstone/src/main/java/com/google/sps/filter/AuthenticateFilter.java
AndrewDPham: Sounds good, I will probably add this to the RequestParser class

mattlgy (approved):

PR #12, opened 2020-08-13
Add headers to request and created logout functionality

AndrewDPham:
Implemented simple way to store ID_TOKEN for scripts to send as headers. Added logout functionality which routes logged out users to index.html.

### DIFF
--- a/capstone/src/main/java/com/google/sps/filter/AuthenticateFilter.java
+++ b/capstone/src/main/java/com/google/sps/filter/AuthenticateFilter.java
@@ -32,7 +32,7 @@ import com.google.sps.data.RequestParser;
 public class AuthenticateFilter implements javax.servlet.Filter {
 
     public void init(FilterConfig filterConfig) throws ServletException{
-
+        
     }
  
     public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
--- a/capstone/src/main/webapp/WEB-INF/web.xml
+++ b/capstone/src/main/webapp/WEB-INF/web.xml
@@ -11,5 +11,6 @@
     <filter-mapping>
         <filter-name>authFilter</filter-name>
         <url-pattern>/dashboard</url-pattern>
+        <url-pattern>/forms</url-pattern>
     </filter-mapping>
 </web-app>
--- a/capstone/src/main/webapp/authenticate.js
+++ b/capstone/src/main/webapp/authenticate.js
@@ -52,7 +52,7 @@ function initClient() {
         scope: SCOPES
     }).then(function () {
         // Set ID_TOKEN and dispatch event
-        ID_TOKEN = gapi.auth2.getAuthInstance().currentUser.get().getAuthResponse().id_token;
+        window.ID_TOKEN = gapi.auth2.getAuthInstance().currentUser.get().getAuthResponse().id_token;
         window.dispatchEvent(new Event('authorized'));
 
         // Sign in and listen for sign-in state changes.
--- a/capstone/src/main/webapp/authenticate.js
+++ b/capstone/src/main/webapp/authenticate.js
@@ -4,7 +4,7 @@ var DISCOVERY_DOCS = ["https://script.googleapis.com/$discovery/rest?version=v1"
 var SCOPES = 'https://www.googleapis.com/auth/forms';
 var ID_TOKEN;
 
-var [dynamicButton, authorizeButton, signoutButton] = createDynamicSigninButton();
+var [dynamicButton, signoutButton] = createDynamicSigninButton();
 document.currentScript.after(createAPIScriptElement());
 window.addEventListener('load', handleClientLoad);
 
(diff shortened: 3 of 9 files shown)


capstone/src/main/webapp/WEB-INF/web.xml
```diff
@@ -11,5 +11,6 @@
     <filter-mapping>
         <filter-name>authFilter</filter-name>
         <url-pattern>/dashboard</url-pattern>
+        <url-pattern>/forms</url-pattern>
```
AndrewDPham, line 14: Apparently /form is a reserved URL. So I changed the mapping of all of the /form to /forms.

capstone/src/main/webapp/authenticate.js
```diff
@@ -2,8 +2,9 @@ var CLIENT_ID = config.CLIENT_ID;
 var API_KEY = config.API_KEY;
 var DISCOVERY_DOCS = ["https://script.googleapis.com/$discovery/rest?version=v1"];
 var SCOPES = 'https://www.googleapis.com/auth/forms';
+var ID_TOKEN;
 
-var [dynamicButton, authorizeButton, signoutButton] = createDynamicSigninButton();
+var [dynamicButton, signoutButton] = createDynamicSigninButton();
```
AndrewDPham, line 7: Thinking about changing this to just a signoutButton now since we don't need a sign in button on internal resources?

AndrewDPham: Meaning that instead of dynamic button, it'll just be signout button.

capstone/src/main/webapp/authenticate.js
```diff
@@ -59,11 +51,12 @@ function initClient() {
         discoveryDocs: DISCOVERY_DOCS,
         scope: SCOPES
     }).then(function () {
+        // Set ID_TOKEN and dispatch event
+        ID_TOKEN = gapi.auth2.getAuthInstance().currentUser.get().getAuthResponse().id_token;
```
mattlgy: I thinking you can do: `window.ID_TOKEN` to make it clear that this is a global variable. In JS, any variable declared in the global context is automatically added to window.

(same for uses of ID_TOKEN below.

mattlgy (changes requested):

mattlgy (approved):

capstone/src/main/webapp/authenticate.js
AndrewDPham: Got it.

capstone/src/main/java/com/google/sps/filter/AuthenticateFilter.java
```diff
@@ -33,7 +33,7 @@
     private static final String AUTH_HEADER_NAME = "id_token";
 
     public void init(FilterConfig filterConfig) throws ServletException{
-
+
```
707090, line 36: Minor thing so not even worth the trouble of updating the PR, but its preferable to group whitespace/formatting updates in their own PR full of meaningless changes so that it is very easy to review and just "yup yup yup, approve". When it is part of content PRs it clutters up the diff

capstone/src/main/webapp/WEB-INF/web.xml
707090: Interesting 🤔 I did not know that

capstone/src/main/webapp/authenticate.js
707090: Seems reasonable if dynamic button is not used anywhere. I suppose you would have other sign in buttons (for example on the home page) just redirect to the login page?

capstone/src/main/webapp/authenticate.js
```diff
@@ -59,11 +51,12 @@ function initClient() {
         discoveryDocs: DISCOVERY_DOCS,
         scope: SCOPES
     }).then(function () {
+        // Set ID_TOKEN and dispatch event
+        window.ID_TOKEN = gapi.auth2.getAuthInstance().currentUser.get().getAuthResponse().id_token;
+        window.dispatchEvent(new Event('authorized'));
```
707090, line 56: Another (preferable) alternative is to add the token to the event itself. Relying on global variables is sometimes worth the convenience, but in this case you can set the id token on the authorized event itself, and then anything which needs the token MUST listen to the authorized event, and cant mistakenly just assume that somehow the global will be set in time before xyz JS is run

capstone/src/main/webapp/authenticate.js
```diff
@@ -12,23 +13,14 @@ window.addEventListener('load', handleClientLoad);
 */
 function createDynamicSigninButton() {
     var dynamicButton = document.createElement("div");
-    var loginButton = dynamicButton.appendChild(createLoginButton());
     var logoutButton = dynamicButton.appendChild(createLogoutButton());
-    return [dynamicButton, loginButton, logoutButton];
```
707090, line 17: At this point dynamic button isnt really dynamic anymore since it is always a logout button right? Because of that, the variable name should be updated, but actually since all it does is append a single child logout button, the places that use it should really just use the logout button and remove the variable altogether

707090 (changes requested):

capstone/src/main/java/com/google/sps/filter/AuthenticateFilter.java
AndrewDPham: Got it. Makes sense.

AndrewDPham: Makes sense. It will feel nice once we have our project done and just removing white space to make it cleaner :D.

PR #13, opened 2020-08-13
Index.html styling

kyleasmith1:
PR focused on index.html styling. Only differences in code between this PR and the last one should be:

index.html
style.css

Screenshot 2020-08-13 at 1 26 39 PM
Screenshot 2020-08-13 at 1 26 25 PM
Screenshot 2020-08-13 at 1 27 39 PM


capstone/src/main/webapp/index.html
```diff
@@ -2,19 +2,71 @@
...
+        <ul class="nav navbar-nav navbar-left">
+            <li><a class="navbar-brand" href="index.html">Rooms</a></li>
+        </ul>
+        <ul class="nav navbar-nav navbar-right">
+            <li><button type="button" class="btn btn-default" onclick="window.location.href='https://github.com/googleinterns/step152-2020'">View On Github</button></li>
+        </ul>
+    </nav>
+ 
+    <div class="container-main container-fluid text-center bg-1 py-1">
+        <h1 class="margin spacing-top">
```
AndrewDPham: I don't think declaring "margin" and "spacing top" as classes to add custom spaces is feasible since we would need to do custom spaces for every object. Bootstrap provides a way to add spacing with "m" or "p" although the spacing is set to only 5 levels (need node.js to have "custom levels"). I'm unsure if there is a better way to add custom spacing in the mean time so would need advice from @mattlgy and @707090 .

capstone/src/main/webapp/index.html
```diff
@@ -2,19 +2,71 @@
...
+    <nav class="navbar navbar-default">
+        <ul class="nav navbar-nav navbar-left">
+            <li><a class="navbar-brand" href="index.html">Rooms</a></li>
+        </ul>
+        <ul class="nav navbar-nav navbar-right">
+            <li><button type="button" class="btn btn-default" onclick="window.location.href='https://github.com/googleinterns/step152-2020'">View On Github</button></li>
+        </ul>
+    </nav>
+ 
+    <div class="container-main container-fluid text-center bg-1 py-1">
```
AndrewDPham: The container-main is color:white but the bg-1 is also color:white with custom spacing. If we go the route of adding custom classes, I think only one custom class should encapsulate the custom styling.

capstone/src/main/webapp/index.html
```diff
@@ -2,19 +2,71 @@
...
+        <h1 class="margin spacing-top"> 
+            <div id="g-theme-blue">The</div> 
+            <div id="g-theme-green">New</div> 
+            Platform 
+            <div id="g-theme-red">for</div> 
+            <div id="g-theme-yellow">Learning</div>
+        </h1> 
+        <img class="img-responsive" src="/assets/soundwave.svg" alt="Sound Wave"> 
+        <p><h4>Serving 314 Users</h4></p>
+        <button type="button" class="btn btn-bg" onclick="window.location.href='login.html'">Join Now</button>
```
AndrewDPham: I don't think setting a custom btn-bg is necessary. We can override the primary buttons to show the colors instead of creating a custom class.

AndrewDPham: There are many type of primary buttons such as btn-primary and btn-outline-primary. If we are going for an "inverted" version of the "Try it out buttons", we can probably use btn-outline-primary here instead.

capstone/src/main/webapp/style.css
```diff
@@ -1,8 +1,135 @@
...
+}
+
 body {
-  margin-left: auto;
-  margin-right: auto;
-  text-align: center;
-  width: 650px;
+    font-family: 'Vollkorn', serif;
+    line-height: 1.8;
+    color: #2c2c2c;
```
AndrewDPham, line 13: If the body has color: #2c2c2c, why do other classes of color properties too? Also do we plan on making the color of the whole website as #2c2c2c?

capstone/src/main/webapp/index.html
kyleasmith1: I was just going to have classes like "margin, margin-lg, margin-sm" that could be our own customizations. That way we don't have to rely on finding the perfect level and we can create the levels ourselves that work with our specific design.

kyleasmith1: But it might make more sense to just use their classes and adjust our site accordingly

capstone/src/main/webapp/index.html
kyleasmith1: Got it

capstone/src/main/webapp/index.html
kyleasmith1: I only created custom classes when I knew that it was specific to that section. That button isn't the same as the "inverted versions" that we used elsewhere and I didn't want to override a pre-made class with something that only the home page will use.

kyleasmith1: I moved it to a bootstrap class that we don't use anywhere else though

capstone/src/main/webapp/style.css
kyleasmith1: Some classes still need the color property even though the default color on the html is #2c2c2c. I thought that #2c2c2c would be our default color and if specific things needed other colors you would create the necessary color attribute (rather than putting #2c2c2c in every class/id).

capstone/src/main/webapp/index.html
```diff
@@ -7,56 +7,56 @@
...
-            Platform 
-            <div id="g-theme-red">for</div> 
-            <div id="g-theme-yellow">Learning</div>
-        </h1> 
-        <img class="img-responsive" src="/assets/soundwave.svg" alt="Sound Wave"> 
-        <p><h4>Serving 314 Users</h4></p>
-        <button type="button" class="btn" onclick="window.location.href='login.html'">Join Now</button>
-        <div class="spacing-bottom"></div>
+    <div class="container-main container-fluid text-center py-1">
+        <div class="py-6">
```
AndrewDPham, line 22: I don't think there is a py-6. The furthest this goes up to is py-5.

mattlgy (approved):

PR #17, opened 2020-08-15
New Data Model for "Rooms"

kyleasmith1:
Added some new classes and updated some old ones. Enjoy


capstone/src/main/java/com/google/sps/data/Lesson.java
```diff
@@ -2,28 +2,25 @@
...
+    protected static final String TITLE_PROPERTY_KEY = "title";
+    protected static final String DESCRIPTION_PROPERTY_KEY = "description";
+    protected static final String DATE_PROPERTY_KEY = "date";
+    protected static final String TAG_LIST_PROPERTY_KEY = "tagList";
+
+    protected static final String TYPE_SURVEY = "survey";
+    protected static final String TYPE_VIDEO = "video";
+    protected static final String TYPE_IMAGE = "image";
+    protected static final String TYPE_CONTENT = "content";
+    protected static final String TYPE_TAG = "tag";
```
AndrewDPham: I actually think this should be public since we would want to reference these variables in servlets and in tests.

capstone/src/main/java/com/google/sps/data/Room.java
```diff
@@ -48,29 +47,71 @@ public void setTitle(String title){
...
+        return (String) this.entity.getProperty(Room.DESCRIPTION_PROPERTY_KEY);
+    }
+
+    public void setDescription(String description) {
+        this.entity.setProperty(Room.DESCRIPTION_PROPERTY_KEY, description);
+    }
+
     @SuppressWarnings("unchecked")
-    public List<Key> getAllFollowers() {
+    public List<Key> getAllFollowers() { // For Andrew: Your data model showed these as actual User objects. Was that on purpose?
```
AndrewDPham: Apologies, they should be keys that represent the user.

capstone/src/main/java/com/google/sps/data/Room.java
```diff
@@ -4,15 +4,15 @@
...
-import com.google.sps.service.DatabaseService;
 
 public class Room {
     public static final String ROOM_ENTITY_NAME = "Room";
     public static final String TITLE_PROPERTY_KEY = "title";
     public static final String DESCRIPTION_PROPERTY_KEY = "description";
     public static final String HOST_PROPERTY_KEY = "host";
     public static final String FOLLOWERS_PROPERTY_KEY = "followers";
-    
+    public static final String TAG_LIST_PROPERTY_KEY = "tagList";
```
AndrewDPham: I think TAGS_PROPERTY_KEY may be better here

capstone/src/main/java/com/google/sps/data/Room.java
kyleasmith1: Got it

capstone/src/main/java/com/google/sps/data/Lesson.java
kyleasmith1: Got it

capstone/src/main/java/com/google/sps/data/Room.java
kyleasmith1: Yup

capstone/src/main/java/com/google/sps/data/Content.java
```diff
@@ -0,0 +1,46 @@
...
+        super.entity.setProperty(Content.URLS_PROPERTY_KEY, urls);
+    }
+
+    public String getContent() {
+        return (String) super.entity.getProperty(Content.CONTENT_PROPERTY_KEY);
+    }
+
+    public List<String> getAllURLs() {
+        return (ArrayList<String>) super.entity.getProperty(Content.URLS_PROPERTY_KEY);
+    }
```
AndrewDPham: I think we need to suppress warnings here.

capstone/src/main/java/com/google/sps/data/Form.java
```diff
@@ -26,8 +26,4 @@ public String getEditURL() {
     public String getURL() {
         return (String) super.entity.getProperty(Form.URL_PROPERTY_KEY);
     }
-
-    public Entity getFormEntity() {
-        return super.entity;
-    }
```
AndrewDPham, line 32: I think we can have this here? Except there shouldn't be a time we down cast a lesson into a form so this actually may not be needed.

capstone/src/main/java/com/google/sps/data/Content.java
kyleasmith1: Think you're right. Must have missed it

capstone/src/main/java/com/google/sps/data/Form.java
kyleasmith1: we have a getEntity() in Lesson.java that should work for all the classes that extend it

kyleasmith1: I could be wrong though

AndrewDPham: Yeah that makes sense. It would not work in the case of Form form = new Form(Blah) but for our case, we would not need to do that in our servlets.

capstone/src/main/java/com/google/sps/data/Lesson.java
```diff
@@ -78,4 +94,12 @@ public void setDescription(String description) {
     public Date getDate() {
         return (Date) this.entity.getProperty(Lesson.DATE_PROPERTY_KEY);
     }
+
+    public Entity getLessonEntity() {
```
707090, line 98: (optional) without any other entity getters, this might as well be called getEntity since we already know it is a lesson from its class.

capstone/src/main/java/com/google/sps/data/Room.java
```diff
@@ -48,29 +66,71 @@ public void setTitle(String title){
...
+        return (String) this.entity.getProperty(Room.DESCRIPTION_PROPERTY_KEY);
+    }
+
+    public void setDescription(String description) {
+        this.entity.setProperty(Room.DESCRIPTION_PROPERTY_KEY, description);
+    }
+
     @SuppressWarnings("unchecked")
-    public List<Key> getAllFollowers() {
+    public List<Key> getAllFollowers() { // For Andrew: Your data model showed these as actual User objects. Was that on purpose?
```
707090: Communication should happen through channels other than checked in code. This comment should be removed (same for all other instances)

capstone/src/main/java/com/google/sps/data/Tag.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final String TAG_MEDIUM_DIFFICULTY = "Medium";
+    private static final String TAG_HARD_DIFFICULTY = "Hard"; 
+
+    private static final String TAG_ONE_MINUTE = "1:00"; 
+    private static final String TAG_TWO_MINUTES = "2:00"; 
+    private static final String TAG_THREE_MINUTES = "3:00"; 
+    private static final String TAG_FOUR_MINUTES = "4:00";
+    private static final String TAG_FIVE_MINUTES = "5:00";
+    private static final String TAG_TEN_MINUTES = "10:00";
+    private static final String TAG_TWENTY_MINUTES = "20:00+";
```
707090: This should be an enum

707090 (changes requested):

capstone/src/main/java/com/google/sps/data/Content.java
```diff
@@ -0,0 +1,46 @@
...
+        super(Lesson.TYPE_CONTENT, title, description);
+        super.entity.setProperty(Content.CONTENT_PROPERTY_KEY, content);
+        super.entity.setProperty(Content.URLS_PROPERTY_KEY, urls);
+    }
+
+    public String getContent() {
+        return (String) super.entity.getProperty(Content.CONTENT_PROPERTY_KEY);
+    }
+
+    public List<String> getAllURLs() {
```
AndrewDPham: Should be Urls.

AndrewDPham: Same goes for the other methods.

capstone/src/main/java/com/google/sps/data/Lesson.java
```diff
@@ -39,8 +36,27 @@ public Lesson(String type, String title, String description) {
         this.entity.setProperty(Lesson.DATE_PROPERTY_KEY, new Date());
     }
 
-    public Entity getLessonEntity() {
-        return this.entity;
+    @SuppressWarnings("unchecked")
+    public void addTag(String tag) {
+        if (this.entity.getProperty(Lesson.TAGS_PROPERTY_KEY) == null) {
```
AndrewDPham, line 41: This should be Room.TAGS_PROPERTY_KEY but we've changed it in a later PR. Feel free to do it here though.

AndrewDPham (changes requested):

capstone/src/main/java/com/google/sps/data/Lesson.java
AndrewDPham: Sounds good. I think for now, I will leave it since many of the other class has a getUserEntity or getRoomEntity. This goes a while back where I suggested we have an interface and anything that has an entity as its instance variable, it could implement getEntity().

AndrewDPham: Maybe in a later PR, we can change the names of all of the getter methods for the other entities.

capstone/src/main/java/com/google/sps/data/Tag.java
AndrewDPham: Got it.

AndrewDPham (approved):

mattlgy (approved):

capstone/src/main/java/com/google/sps/data/Lesson.java
707090: Yep, grouping all of these together into one PR as opposed to piecemeal is a great idea (if we have time)

capstone/src/main/java/com/google/sps/data/Form.java
707090: NB it would actually work fine. The lesson entity getter returns the exact same (super.) entity that this method is, so whether you created it as a form or just got it as a lesson that happens to be a form, they are editing the same underlying form

capstone/src/main/java/com/google/sps/data/Tag.java
```diff
@@ -0,0 +1,104 @@
...
+    public static List<String> all() {
+        List<String> tags = new ArrayList<>(Tag.CATEGORY_TAGS);
+        tags.addAll(Tag.DIFFICULTY_TAGS);
+        tags.addAll(Tag.TIME_TAGS);
+        return tags;
+    }
+
+    public String getTag() {
+        return this.tag;
+    }
```
707090, line 103: These should be lists of Tags, not strings. Strings may coincidentally be what identifys a tag on interest vectors, but ultimately these collections should be full tag lists. This is because you may want to expand the idea of what a tag is, such as to add tag "weight" or "hotness" or any other thing. Also, for the all() function, that constructs a brand new list every time its called. But ultimately I dont think you need this, since anyone who would call this (after it becomes list of Tags instead of strings) can instead call Tag.values()

707090 (approved):

capstone/src/main/java/com/google/sps/data/Tag.java
AndrewDPham: Sounds good! Yeah I was thinking of whether to make it a collection of Strings or Tags. My original thought process was to call it string since we've only used string in out project and to change it in the future or add another static variable that is a collection of tags in a future PR if we ever needed them.

PR #18, opened 2020-08-17
Serialization

AndrewDPham:
Create serialization and deserialization of lessons methods. Add tests to each method.


capstone/src/main/java/com/google/sps/data/Lesson.java
```diff
@@ -95,6 +100,28 @@ public Date getDate() {
...
     }
 
+    public static Lesson deserializeJson(String json) throws IOException {
+        JsonObject jobject = JsonParser.parseString(json).getAsJsonObject();
+
+        switch(jobject.get(Lesson.TYPE_PROPERTY_KEY).getAsString()) {
+            case Lesson.TYPE_FORM:
+                return new Form(jobject.get(Lesson.TITLE_PROPERTY_KEY).getAsString(), 
+                    jobject.get(Lesson.DESCRIPTION_PROPERTY_KEY).getAsString(), jobject.get(Form.EDIT_URL_PROPERTY_KEY).getAsString(),
+                    jobject.get(Form.URL_PROPERTY_KEY).getAsString());
```
AndrewDPham: There is only the case for forms right now. I can add the cases for the other types of lessons (video, image, content) once I know how to do tests more efficiently.

capstone/src/main/java/com/google/sps/data/Lesson.java
```diff
@@ -95,6 +100,28 @@ public Date getDate() {
...
+            default:
+                throw new IOException(); 
+        }
+    }
+
+    public static Lesson serializeJson(Entity entity) throws IOException {
+        switch((String) entity.getProperty(Lesson.TYPE_PROPERTY_KEY)) {
+            case Lesson.TYPE_FORM:
```
AndrewDPham: Same as deserializeJson

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    @Test
+    public void deserializeJsonGeneralTest() {
+        JsonObject jsonObject = new JsonObject();
+        jsonObject.addProperty(Lesson.TYPE_PROPERTY_KEY, Lesson.TYPE_FORM);
+        jsonObject.addProperty(Lesson.TITLE_PROPERTY_KEY, LessonDeserializationTest.TEST_TITLE);
+        jsonObject.addProperty(Lesson.DESCRIPTION_PROPERTY_KEY, LessonDeserializationTest.TEST_DESCRIPTION);
+        jsonObject.addProperty(Form.EDIT_URL_PROPERTY_KEY, LessonDeserializationTest.TEST_URL);
+        jsonObject.addProperty(Form.URL_PROPERTY_KEY, LessonDeserializationTest.TEST_EDIT_URL);
```
AndrewDPham: It seems I keep creating a different object to use per test and these objects differ slightly (i.e. Missing field, wrong field, extra field). Is there a more efficient way than doing this? Was thinking of creating a variable in setUp() but each object differs slightly per test.

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
```diff
@@ -0,0 +1,219 @@
...
+            Lesson lesson = Lesson.deserializeJson(jsonObject.toString());
+
+            Assert.assertEquals(lesson.getType(), Lesson.TYPE_FORM);
+            Assert.assertEquals(lesson.getTitle(), LessonDeserializationTest.TEST_TITLE);
+            Assert.assertEquals(lesson.getDescription(), LessonDeserializationTest.TEST_DESCRIPTION);
+            Assert.assertEquals(lesson.getIsDraft(), false);
+            Assert.assertTrue(lesson instanceof Form);
+            Form form = (Form) lesson; 
+            Assert.assertEquals(form.getEditUrl(), LessonDeserializationTest.TEST_URL);
+            Assert.assertEquals(form.getUrl(), LessonDeserializationTest.TEST_EDIT_URL);
```
AndrewDPham: Small mistake, the TEST_URL and TEST_EDIT_URL should be switched. Same goes for the other tests. Will fix in next push.

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
```diff
@@ -0,0 +1,219 @@
...
+        jsonObject.addProperty(Form.URL_PROPERTY_KEY, LessonDeserializationTest.TEST_EDIT_URL);
+
+        try{ 
+            Lesson lesson = Lesson.deserializeJson(jsonObject.toString());
+
+            Assert.assertEquals(lesson.getType(), Lesson.TYPE_FORM);
+            Assert.assertEquals(lesson.getTitle(), LessonDeserializationTest.TEST_TITLE);
+            Assert.assertEquals(lesson.getDescription(), LessonDeserializationTest.TEST_DESCRIPTION);
+            Assert.assertEquals(lesson.getIsDraft(), false);
+            Assert.assertTrue(lesson instanceof Form);
```
AndrewDPham: I could move some of the blocks of code out of the try block too unless it seems more logical to put it here?

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
```diff
@@ -0,0 +1,219 @@
...
+        jsonObject.addProperty(Form.URL_PROPERTY_KEY, LessonDeserializationTest.TEST_EDIT_URL);
+
+        try{ 
+            Lesson lesson = Lesson.deserializeJson(jsonObject.toString());
+            Assert.fail();
+        } catch(NullPointerException e) {
+            return;
+        } catch(Exception e) {
+            Assert.fail();
+        }
```
AndrewDPham: This is a test to see deserializeJson fail and throw a NullPointerException (when it should). Was wondering if this is the right way to catch the error we are looking for unless there is another "best practice" way.

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
```diff
@@ -0,0 +1,219 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.sps.data.Form;
+import com.google.sps.data.Lesson;
+import java.io.IOException;
+import com.google.appengine.api.datastore.Entity;
+
+@RunWith(JUnit4.class)
+public final class LessonDeserializationTest {
+
```
AndrewDPham: Created two new files when originally had one LessonTest class to test both serialization and deserialization methods. I believe unit testing means creating a class per method?

707090: Nope, generally you use a class per class, and you just put multiple tests per method next to each other. So:
```
class Foo {
    public String bar() {...}
    public String baz() {...}
}
class FooTest {
    public barTestOne() {...}
    public barTestTwo() {...}
    public bazTestOne() {...}
}
```

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
707090: You can set up baseObject in setUp(), then in each test do baseObject.addProperty, baseObject.removeProperty, etc to create a specific test case. But also copy-pasting is not so bad in tests because it puts everything in one place that you can just read the test from top to bottom. Since tests arent depended on by any code, its not like two values can "get out of sync" or some of the other problems that code dupe has in normal code, so ... 🤷‍♂️ could go either way on this one

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
707090: Nice catch. Did the tests pass anyway? Before each push I usually run tests ("usually"... :kappa:, that is: when Im not being lazy) just to make sure I dont commit or push anything that doesnt pass all tests

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
707090: @ Test(expected = NullPointerException.class)

then just do it without any catches

707090: You might also want to wrap this (and the other ones) in a better exception name. So in the parse you would catch NullPointer or IOException and throw an https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/IllegalArgumentException.html (or any other one that fits the bill and is a little more descriptive of what is wrong)

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
```diff
@@ -0,0 +1,219 @@
...
+            Assert.assertEquals(lesson.getDescription(), LessonDeserializationTest.TEST_DESCRIPTION);
+            Assert.assertEquals(lesson.getIsDraft(), false);
+            Assert.assertTrue(lesson instanceof Form);
+            Form form = (Form) lesson; 
+            Assert.assertEquals(form.getEditUrl(), LessonDeserializationTest.TEST_URL);
+            Assert.assertEquals(form.getUrl(), LessonDeserializationTest.TEST_EDIT_URL);
+            
+        } catch (IOException e) {
+            Assert.fail("IOException Thrown");
+        }
```
707090: I dont think you need this? Im pretty sure the JUnit runner automatically does a try-catch around each test and will just naturally fail the test if an uncaught exception is thrown

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
707090: See below

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
AndrewDPham: Yeah, it passed. This is cause I also compared the editUrl to TEST_URL in the assertEquals method.

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
AndrewDPham: Got it. Makes a lot more sense

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
AndrewDPham: Got it. I decided to leave it as creating the object per test.

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
AndrewDPham: I see, I threw an IllegalArgumentException instead of IOException.

capstone/src/test/java/com/google/sps/LessonDeserializationTest.java
AndrewDPham: Got it.

capstone/src/main/java/com/google/sps/data/Lesson.java
```diff
@@ -95,6 +102,33 @@ public Date getDate() {
         return (Date) this.entity.getProperty(Lesson.DATE_PROPERTY_KEY);
     }
 
+    public static Lesson deserializeJson(String json) throws IllegalArgumentException {
+        JsonObject jobject = JsonParser.parseString(json).getAsJsonObject();
+        
+        switch(jobject.get(Lesson.TYPE_PROPERTY_KEY).getAsString()) {
```
AndrewDPham: This throws a NullPointerException (when there is no field for type) which is caught in my tests. I'm thinking a more informative error is IllegalArgumentException and so I can do:
String type;
try {
    type = jobject.get(Lesson.TYPE_PROPERTY_KEY).getAsString());
} catch (NullPointerException e){
    throw new IllegalArgumentException();
}

Or is this not necessary since we are catching an unchecked exception?

capstone/src/main/java/com/google/sps/service/DatabaseService.java
```diff
@@ -19,8 +23,20 @@ public static Room getRoom(Key key) throws EntityNotFoundException {
         return new Room(DatastoreServiceFactory.getDatastoreService().get(key));  
     }
 
-    public static Form getForm(Key key) throws EntityNotFoundException {
-        return new Form(DatastoreServiceFactory.getDatastoreService().get(key));
+    public static Lesson getLesson(Key key) throws EntityNotFoundException {
+        Entity entity = DatastoreServiceFactory.getDatastoreService().get(key);
+        switch((String) entity.getProperty(Lesson.TYPE_PROPERTY_KEY)) {
```
AndrewDPham: Same as deserializeJson

capstone/src/test/java/com/google/sps/LessonTest.java
```diff
@@ -28,4 +49,150 @@ public void setUp() {
...
+        JsonArray jsonArray = new JsonArray();
+        jsonArray.add(LessonTest.TEST_URL);
+        jsonArray.add(LessonTest.TEST_EDIT_URL);
+        jsonObject.add(Content.URLS_PROPERTY_KEY, (JsonElement) jsonArray);
+
+        ArrayList<String> testList = new ArrayList<>();
+        Iterator<JsonElement> listIterator = jsonArray.iterator();
+        while(listIterator.hasNext()){
+            testList.add(listIterator.next().getAsString());
+        }
```
AndrewDPham, line 129: I found this to be the best way to get replicate a collection of URLs. For future improvement, I could put the construction of the jsonarray in setUp() but decided to leave it here since this is the only test that use it (can add to setUp() in future PR when we test more).

AndrewDPham: Also is there another "better" way to get a jsonArray to an array of type String to compare in assertArrayEquals()?

capstone/src/test/java/com/google/sps/DatabaseServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+        try{ 
+            Lesson lesson = DatabaseService.getLesson(results.asSingleEntity().getKey());
+
+            Assert.assertEquals(lesson.getType(), Lesson.TYPE_FORM);
+            Assert.assertEquals(lesson.getTitle(), DatabaseServiceTest.TEST_TITLE);
+            Assert.assertEquals(lesson.getDescription(), DatabaseServiceTest.TEST_DESCRIPTION);
+            Assert.assertTrue(lesson instanceof Form);
+            Form form = (Form) lesson; 
+            Assert.assertEquals(form.getEditUrl(), DatabaseServiceTest.TEST_EDIT_URL);
+            Assert.assertEquals(form.getUrl(), DatabaseServiceTest.TEST_URL);
```
AndrewDPham: Unsure whether I can leave this in the try block (for "logic") or should I leave it out and place it after catch? EntityNotFoundException is a checked exception.

capstone/src/test/java/com/google/sps/LessonTest.java
```diff
@@ -28,4 +49,150 @@ public void setUp() {
...
+        Lesson lesson = Lesson.deserializeJson(jsonObject.toString());
+
+        Assert.assertEquals(lesson.getType(), Lesson.TYPE_FORM);
+        Assert.assertEquals(lesson.getTitle(), LessonTest.TEST_TITLE);
+        Assert.assertEquals(lesson.getDescription(), LessonTest.TEST_DESCRIPTION);
+        Assert.assertEquals(lesson.getIsDraft(), false);
+        Assert.assertTrue(lesson instanceof Form);
+        Form form = (Form) lesson; 
+        Assert.assertEquals(form.getEditUrl(), LessonTest.TEST_EDIT_URL);
+        Assert.assertEquals(form.getUrl(), LessonTest.TEST_URL);
```
AndrewDPham, line 71: Decided to type cast for each different type (video, lesson, form, content) so that if more fields are added to each one, we wouldn't have to type cast inline to call the method.

capstone/src/test/java/com/google/sps/LessonTest.java
```diff
@@ -28,4 +49,150 @@ public void setUp() {
...
+    }
```
AndrewDPham, line 197: Decided to leave the other tests for each type since I didn't want to let this PR be big. I could add it in a future PR along with servlet testing (I got it to work I think).

AndrewDPham: Wrong area to write this.

capstone/src/test/java/com/google/sps/DatabaseServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+        DatabaseService.save(entity);
+        Query query = new Query(Lesson.LESSON_ENTITY_NAME);
+        PreparedQuery results = datastore.prepare(query);
+
+        try{ 
+            Lesson lesson = DatabaseService.getLesson(results.asSingleEntity().getKey());
+        } catch (EntityNotFoundException e) {
+            Assert.fail("EntityNotFoundException Thrown");
+        }
+    }
```
AndrewDPham, line 106: Decided to leave the other tests for each type since I didn't want to let this PR be big. I could add it in a future PR along with servlet testing (I got it to work I think).

capstone/src/main/java/com/google/sps/data/Lesson.java
707090: Yep, thats exactly what I would do

capstone/src/test/java/com/google/sps/LessonTest.java
707090: I dont know that there is one that is much "better". You could use streams and that might be _slightly_ more readable, but if you are not familiar with the API, then its not worth changing. Java is known for being a somewhat verbose language, so... it taking this much code is somewhat expected.

I think its the right call to leave it in this test, and even potentially copy-paste it if multiple tests use it. setUp is more meant for things which are completely irrelevant to the test, like... idk mocking out some dependencies or something

capstone/src/main/java/com/google/sps/service/DatabaseService.java
707090: Temporal comments should be made in Github (like "Im going to fix this in the next PR" or this hack should be unecessary in a week" etc), and content comments should be made as an actual code comment so that they can be seen forever

capstone/src/test/java/com/google/sps/DatabaseServiceTest.java
707090: You should be able to have ll of this at the top level, and add a throws decl to the function?

707090 (changes requested):

capstone/src/test/java/com/google/sps/LessonTest.java
AndrewDPham: Alrighty. Got it

capstone/src/test/java/com/google/sps/DatabaseServiceTest.java
AndrewDPham: Ahh I think I misunderstood the answer on the previous comment that dealt with this. Got it.

capstone/src/main/java/com/google/sps/service/DatabaseService.java
AndrewDPham: For sure.

capstone/src/main/java/com/google/sps/data/Lesson.java
```diff
@@ -104,28 +104,40 @@ public Date getDate() {
...
+                    List<String> urlList = new Gson().fromJson(listJson, listType);
+
+                    return new Content(jobject.get(Lesson.TITLE_PROPERTY_KEY).getAsString(), 
+                        jobject.get(Lesson.DESCRIPTION_PROPERTY_KEY).getAsString(), jobject.get(Content.CONTENT_PROPERTY_KEY).getAsString(),
+                        urlList);
+                default:
+                    throw new IllegalArgumentException(); 
+            }
+        } catch (NullPointerException e) {
+            throw new IllegalArgumentException();
```
707090, line 140: (optional/can be done in a later PR) these dont need to be in separate try blocks. They are both catching the same exception and are immediately consecutive, so they should be in a single block to make it easier to read

707090 (approved):

capstone/src/main/java/com/google/sps/data/Lesson.java
AndrewDPham: Got it. My thought process was that they are catching exceptions at different times so they would each need their own try and catch block. This makes more sense though. I plan to keep note and fix in a later PR.

mattlgy (approved):

PR #20, opened 2020-08-18
Add get lesson functionality

kyleasmith1:
Added a way to get lessons within specific rooms.

Future updates:

UI, Removing hard-coded values


capstone/src/main/java/com/google/sps/data/Room.java
```diff
@@ -98,7 +98,7 @@ public void removeFollower(User follower) {
     }
 
     @SuppressWarnings("unchecked")
-    public List<Key> getAllLessons() { // For Andrew: Your data model showed these as actual Lesson objects. Was that on purpose?
```
707090: 😄  This kind of comment really shouldnt ever be committed, just send him a ping or something

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
```diff
@@ -12,10 +26,53 @@
...
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery roomResults = datastore.prepare(new Query(Room.ROOM_ENTITY_NAME));
+        Room room = null;
+
+        for(Entity entity : roomResults.asIterable()) {
+            if (new Room(entity).getRoomKey().getId() == Long.parseLong(request.getParameter("room_id"))) {
+                room = new Room(entity);
+            }
+        }
```
707090: When you get the room like this, you are querying for ALL rooms, then finding the right one in a for loop. That doesnt allow indexes to work their magic, so as the database grows, this would become incredibly slow. Instead you should set a query filter

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
```diff
@@ -12,10 +26,53 @@
...
+    public List<Lesson> getLessons(Room room) {
+        List<Lesson> lessons = new ArrayList<>();
+    
+        try {
+            for(Key key : room.getAllLessons()) {
+                lessons.add(DatabaseService.getLesson(key));
+            }
+        }
+        catch (EntityNotFoundException e) {
+            System.out.println("Lesson entities don't exist.");
```
707090: System.err.println

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -36,7 +36,8 @@ function createRoomDivElement(room) {
                     <div class="card-text small-text">Followers: Infinite</div>
                     <div class="card-text small-text">Tag(s): </div>
                     <div class="small-spacing-bottom"></div>
-                    <button type="button" class="btn btn-default" onclick="window.location.href='lesson.html'">View</button>
+                    <button type="button" class="btn btn-default" 
+                    onclick="window.location.href='lesson.html?room_id='+${room.entity.key.id}">View</button>
```
707090, line 37: Same injection problem as before. I remember you said you were going to take care of it in a future PR, just noting.

capstone/src/main/webapp/lesson-loader.js
```diff
@@ -58,11 +59,47 @@ function createImage(title, description, url) {
...
+            lessonElement.appendChild(createLessonDivElement(lesson));
+        };
+    });
+}
+
+function getRoomId() {
+    var site = window.location.href;
+    const loc = site.indexOf("=") + 1;
+    var roomId = site.substring(loc);
+    return roomId;
```
707090: new URL(window.location.href).searchParams.get("room_id")

707090 (changes requested):

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
kyleasmith1: Got it

capstone/src/main/webapp/dashboard-loader.js
kyleasmith1: Yup good note.

capstone/src/main/webapp/lesson-loader.js
kyleasmith1: Much better thanks

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
kyleasmith1: I don't know how I would check to see if the Room key is the same as the query parameter with no static variables set to "roomId" in the Room class. If you check out LoginServlet it was easy there because the User stores his Id but I didn't know if there was a special way to do this when an object doesn't have the filter specific field?

kyleasmith1: Fixed

capstone/src/main/java/com/google/sps/service/FilterService.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory.Builder;
```
707090: Usually for the builder patterns you dont import the full path because that would mean if you also wanted to use something like Query.Builder they would conflict. So you just import up to the base class (KeyFactory) and then when you reference it you can use Key key = new KeyFactory.Builder(...)...

capstone/src/main/java/com/google/sps/service/FilterService.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class FilterService {
+
+    private static final String KEY_PROPERTY = "__key__";
+
+    public static Entity getEntity(String entity, Long id) {
```
707090: entity could be named something which differentiates it from sounding like an Entity (which it clearly isnt since its a String, but still isnt clear what it is supposed to be from just the signature). Something like entity_name is easier to understand from just the signature.

capstone/src/main/java/com/google/sps/service/FilterService.java
```diff
@@ -0,0 +1,35 @@
...
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Key key = new Builder(entity, id).getKey();
+        Filter filter = new FilterPredicate(KEY_PROPERTY, FilterOperator.EQUAL, key);
+        Query query = new Query(entity).setFilter(filter);
+        PreparedQuery result = datastore.prepare(query);
+        return result.asSingleEntity();
+    }
+
+    public static PreparedQuery getQuery(String entity, String property, Key key) {
```
707090: getQuery doesnt describe what the method does? A query of what?

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
```diff
@@ -71,7 +54,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
             }
         }
         catch (EntityNotFoundException e) {
-            System.out.println("Lesson entities don't exist.");
+            System.err.println("Lesson entities don't exist.");
```
707090: Printing additional context information helps when you are trying to debug a problem by looking at logs. So you could also print out the room ID which was expected to exist and maybe the room name or something

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
```diff
@@ -30,11 +28,8 @@
 
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         Key userKey = ((User) request.getAttribute(User.USER_ENTITY_NAME)).getUserKey();
-        Filter roomFilter = new FilterPredicate(Room.HOST_PROPERTY_KEY, FilterOperator.EQUAL, userKey);
-        Query query = new Query(Room.ROOM_ENTITY_NAME).setFilter(roomFilter);
-        PreparedQuery results = datastore.prepare(query);
+        PreparedQuery results = FilterService.getQuery(Room.ROOM_ENTITY_NAME, Room.HOST_PROPERTY_KEY, userKey);
```
707090: This is really good code de-duplication. I wonder if the helper method currently called "getQuery" could also include turning it into a list like happens below? It seems that is also part of the repeating pattern

707090 (changes requested):

capstone/src/main/java/com/google/sps/service/FilterService.java
kyleasmith1: got it

capstone/src/main/java/com/google/sps/servlets/DashboardHandlerServlet.java
kyleasmith1: I would have moved the list into the helper function but Idk how I would get the correct object type to pass in as a parameter? Because the type that the list holds has to be specified, I don't know how I would pass that information over

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
```diff
@@ -12,10 +24,38 @@
...
+    public List<Lesson> getLessons(Room room) {
+        List<Lesson> lessons = new ArrayList<>();
+    
+        try {
+            for(Key key : room.getAllLessons()) {
+                lessons.add(DatabaseService.getLesson(key));
+            }
+        }
+        catch (EntityNotFoundException e) {
+            System.err.println("Lesson entities don't exist. Room: " + room.getKey() + " Title: " + room.getTitle());
```
AndrewDPham: This throws a compile error. Please compile before pushing :). I think it should be room.getRoomKey();

capstone/src/main/java/com/google/sps/service/FilterService.java
```diff
@@ -10,26 +10,27 @@
...
+        Key key = new KeyFactory.Builder(entity_name, id).getKey();
         Filter filter = new FilterPredicate(KEY_PROPERTY, FilterOperator.EQUAL, key);
-        Query query = new Query(entity).setFilter(filter);
+        Query query = new Query(entity_name).setFilter(filter);
         PreparedQuery result = datastore.prepare(query);
         return result.asSingleEntity();
     }
 
-    public static PreparedQuery getQuery(String entity, String property, Key key) {
+    public static PreparedQuery getQueryOfKey(String entity_name, String property, Key key) {
```
707090: This is still not particularly informative. Maybe call it getEntityQueryByKeyedProperty

707090 (commented):
You can pass it back with method level generics, so that the method specifies what kind of list it wants back. https://docs.oracle.com/javase/tutorial/extra/generics/methods.html

AndrewDPham (changes requested):

707090 (approved):

PR #21, opened 2020-08-20
Follow

kyleasmith1:
Added join/unjoin functionality

### DIFF
--- a/capstone/src/main/java/com/google/sps/servlets/JoinServlet.java
+++ b/capstone/src/main/java/com/google/sps/servlets/JoinServlet.java
@@ -1,0 +1,61 @@
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import com.google.sps.data.Lesson;
+import com.google.sps.data.Room;
+import com.google.sps.data.User;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.EntityNotFoundException;
+import java.util.List;
+import java.util.ArrayList;
+import com.google.gson.Gson;
+import com.google.sps.data.Lesson;
+import com.google.sps.data.Room;
+import com.google.sps.data.RequestParser;
+import com.google.sps.service.DatabaseService;
+import com.google.sps.data.RequestParser;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/join")
+public class JoinServlet extends HttpServlet {
+
+    private static final String JOIN_PROPERTY_KEY = "join";
+    private static final String UNJOIN_PROPERTY_KEY = "unjoin";
+    private static final String ROOM_ID_PROPERTY_KEY = "room_id";
+    private static final String ACTION_PROPERTY_KEY = "action";
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        User user = (User) request.getAttribute(User.USER_ENTITY_NAME);
+        PreparedQuery roomResults = datastore.prepare(new Query(Room.ROOM_ENTITY_NAME));
+        Room room = null;
+
+        for(Entity entity : roomResults.asIterable()) {
+            if (new Room(entity).getRoomKey().getId() == Long.parseLong(request.getParameter(JoinServlet.ROOM_ID_PROPERTY_KEY))) {
+                room = new Room(entity);
+            }
+        }
+
+        if (request.getParameter(JoinServlet.ACTION_PROPERTY_KEY).equals(JoinServlet.JOIN_PROPERTY_KEY) && !(room.getAllFollowers().contains(user.getUserKey()))) {
+            room.addFollower(user);
+        } 
+        
+        if (request.getParameter(JoinServlet.ACTION_PROPERTY_KEY).equals(JoinServlet.UNJOIN_PROPERTY_KEY)) {
+            room.removeFollower(user);
+        }
+
+        DatabaseService.save(room.getRoomEntity());
+
+        System.out.println(room.getRoomEntity());
+        response.setStatus(HttpServletResponse.SC_OK);
+    }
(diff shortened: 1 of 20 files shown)


capstone/src/main/java/com/google/sps/servlets/JoinServlet.java
```diff
@@ -0,0 +1,59 @@
...
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        User user = (User) request.getAttribute(User.USER_ENTITY_NAME);
+        PreparedQuery roomResults = datastore.prepare(new Query(Room.ROOM_ENTITY_NAME));
+        Room room = null;
+
+        for(Entity entity : roomResults.asIterable()) {
+            if (new Room(entity).getRoomKey().getId() == Long.parseLong(request.getParameter(JoinServlet.ROOM_ID_PROPERTY_KEY))) {
+                room = new Room(entity);
+            }
+        }
```
707090: This should not query every single room, it should use a query filter

capstone/src/main/java/com/google/sps/servlets/JoinServlet.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/join")
+public class JoinServlet extends HttpServlet {
+
+    private static final String JOIN_PROPERTY_KEY = "join";
+    private static final String UNJOIN_PROPERTY_KEY = "unjoin";
+    private static final String ROOM_ID_PROPERTY_KEY = "room_id";
+    private static final String ACTION_PROPERTY_KEY = "action";
```
707090: These are not property keys. The last one is a query parameter, the room_id one is a property key, and the first two are just action values. Since the join action can only have one of a set number of possible values, use an enum

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -28,15 +28,20 @@ function createRoomDivElement(room) {
...
+                    <div class="btn btn-default" id="room-link"></button>
                 </div>
             </div>
             `, "text/html");
+    
+    doc.getElementById("room-title").innerText = room.entity.propertyMap.title;
+    doc.getElementById("room-link").addEventListener("click", function() {
+        window.location.href = "lesson.html?room_id=" + room.entity.key.id;
+    });
+    doc.getElementById("room-link").innerText = "View";
```
707090: This sanitization is not related to following a room and should be part of a separate PR for handling sanitizaiton

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -28,15 +28,20 @@ function createRoomDivElement(room) {
...
                 <div class="card-body text-center">
-                    <h5 class="card-title">${room.entity.propertyMap.title}</h5>
+                    <h5 class="card-title" id="room-title"></h5>
                     <a class="card-text small-text" href="#">Kyle Smith</a>
                     <div class="card-text small-text">Followers: Infinite</div>
                     <div class="card-text small-text">Tag(s): </div>
                     <div class="small-spacing-bottom"></div>
-                    <button type="button" class="btn btn-default" 
-                    onclick="window.location.href='lesson.html?room_id='+${room.entity.key.id}">View</button>
+                    <div class="btn btn-default" id="room-link"></button>
```
707090: \<div ...>\</button>? This doesnt freak the browser out? Most browsers are pretty forgiving, but Im surprised this works. In any case, this should be an <a> anyway (per below, changed in a separate PR)

capstone/src/main/webapp/lesson-loader.js
```diff
@@ -6,6 +6,8 @@ const VIDEO = "video";
 const CONTENT = "content";
 const IMAGE = "image";
 
+const queryString = "/lesson?room_id=" + getRoomId() + "&action=" + getJoinStatus();
```
707090: Terminology-wise, the query string in a url is purely the part after the '?' (and before the '#' if there is one), not the whole path. I would just call this like lessonPath or something. Good call pulling this out though

capstone/src/main/webapp/lesson-loader.js
```diff
@@ -78,18 +80,36 @@ function getLessons() {
     });
 }
 
+function joinRoom() {
+    console.log("Room joined!");
```
707090: Debug lines should be removed before PRs

capstone/src/main/webapp/lesson-loader.js
```diff
@@ -78,18 +80,36 @@ function getLessons() {
...
+    return fetch("/join?room_id=" + getRoomId() + "&action=unjoin",
+    {method: "POST", headers: new Headers({ID_TOKEN})});
+}
+
 function getRoomId() {
     return new URL(window.location.href).searchParams.get("room_id");
 }
 
-// Will update in UI PR
+function getJoinStatus() {
```
707090: This could be called getJoinAction, since "unjoining" is not a status

capstone/src/main/webapp/lesson-loader.js
707090: To make this more robust, I would use the JS classes around URLs (this includes for all requests in the app). You can write this as:

```
lessonEndpoint = new URL("/lesson", window.location.href);
lessonEndpoint.searchParams.set(ROOM_ID, getRoomId());
lessonEndpoint.searchParams.set(ACTION, getJoinAction());
```
then later when you want to use it you can fetch this instead of a string `fetch(locationEndpoint, {method...})`

707090 (changes requested):
Unjoin can also probably be called leave?

capstone/src/main/java/com/google/sps/servlets/JoinServlet.java
kyleasmith1: gotcha

capstone/src/main/webapp/dashboard-loader.js
kyleasmith1: yea sorry I should've made a separate PR

capstone/src/main/webapp/dashboard-loader.js
kyleasmith1: What should this have been? Like why did you expect it to freak the browser out?

capstone/src/main/webapp/lesson-loader.js
kyleasmith1: gotcha that works perfectly

capstone/src/main/webapp/dashboard-loader.js
707090: \<div>\</button>

kyleasmith1: I didn't notice and the browser didn't catch it I guess. I'll change it now so I can merge

707090 (approved):

PR #22, opened 2020-08-20
Tag update

AndrewDPham:
Update tags to show better categories. Add constructUserVectorMap() that represents the vectorMap of a user.


capstone/src/main/java/com/google/sps/data/Tag.java
```diff
@@ -98,4 +93,30 @@ private Tag() {
...
+        vector.put(Tag.TAG_SPACE, 0);
+        vector.put(Tag.TAG_CAREER, 0);
+        vector.put(Tag.TAG_ARCHITECTURE, 0);
+        vector.put(Tag.TAG_ART, 0);
+        vector.put(Tag.TAG_DANCING, 0);
+        vector.put(Tag.TAG_CHALLENGE, 0);
+        vector.put(Tag.TAG_DIETING, 0);
+        vector.put(Tag.TAG_ADVENTURE, 0);
+        vector.put(Tag.TAG_MEDITATION, 0);
+        vector.put(Tag.TAG_NEWS, 0);
```
707090: This should be a for loop. As it stands, you have to remember to make a change here if you add a tag. (Also, per a previous PR comment, this will become an enum anyway instead of a bunch of constants)

capstone/src/test/java/com/google/sps/LessonTest.java
```diff
@@ -31,8 +31,7 @@
 
     //Helper to set local datastore service
     private final LocalServiceTestHelper helper =
-      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig()
-          .setDefaultHighRepJobPolicyUnappliedJobPercentage(100));
+      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
```
707090: Since this is a completely unrelated change it should not be part of this PR. Its fine to create a 1 line PR if this is not related to anything else at all, but having it piggyback as part of another CL can cause problems if for example this CL caused a bug and needed to be rolled back

707090 (changes requested):

707090:
The fact that this PR was nice and tiny made it really easy to review 👍

capstone/src/main/java/com/google/sps/data/Tag.java
AndrewDPham: Got it.

capstone/src/test/java/com/google/sps/LessonTest.java
AndrewDPham: Sounds good. I just thought having a 1 line PR seems really minuscule but this makes a lot more sense.

capstone/src/main/java/com/google/sps/data/Tag.java
```diff
@@ -101,4 +101,14 @@ private Tag(String tag) {
...
+
+    public static final HashMap<String, Integer> constructUserVectorMap() {
+        HashMap<String, Integer> vector = new HashMap<>();
+
+        for(String tag : CATEGORY_TAGS) {
+            vector.put(tag, 0);
+        }
+        
+        return vector;
+    }
```
707090: (optional) I would recommend not having this. Instead, just create a new hashmap, and wherever you are using this, do checks to see if a key is already present instead of just getting it and knowing its there. Bugs could crop up if for example this only pre-fills categories but later down the road the reccomender takes into account length tags for example. Also, its unnecessary work to build up this map if for example only 2 categories are interacted with

707090 (approved):

mattlgy (approved):

AndrewDPham:
Closing to create new PR.

PR #23, opened 2020-08-20
Recommender algo

AndrewDPham:
Create populateUserTagMap() which creates the HashMap to represent the User vector. Create userTagMap for user being recommended and a Hashmap of otherUser vectors that will aid in recommending. Add tests to tests populateUserTagMap.

### DIFF
--- a/capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
+++ b/capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
@@ -34,6 +34,7 @@ public class RecommenderAlgorithm {
             otherUserVectorMap.put(otherUser, otherUserTagMap);
         }
  
+        // Algorithm is work in progress. Have yet to normalize and calculate relationship.
         return null;
     }
  
--- a/capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
+++ b/capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
@@ -1,0 +1,58 @@
+package com.google.sps.data;
+ 
+import com.google.sps.data.Room;
+import com.google.sps.data.User;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.Entity;
+import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.List;
+import java.lang.Iterable;
+ 
+public class RecommenderAlgorithm {
+ 
+    public static ArrayList<Room> recommendRooms(User user) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        HashMap<String, Integer> userTagMap = RecommenderAlgorithm.populateUserTagMap(user);
+        
+        Query otherUserQuery = new Query(User.USER_ENTITY_NAME);
+        PreparedQuery otherUserResults = datastore.prepare(otherUserQuery);
+ 
+        User otherUser = null;
+        HashMap<String, Integer> otherUserTagMap = null;
+        HashMap<User, HashMap<String,Integer>> otherUserVectorMap = new HashMap<User, HashMap<String,Integer>>();
+        for(Entity otherUserEntity : otherUserResults.asIterable()) {
+            otherUser = new User(otherUserEntity);
+            otherUserTagMap = RecommenderAlgorithm.populateUserTagMap(otherUser);
+            otherUserVectorMap.put(otherUser, otherUserTagMap);
+        }
+ 
+        return null;
+    }
+ 
+    public static HashMap<String, Integer> populateUserTagMap(User user) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Filter userRoomFilter = new FilterPredicate(Room.FOLLOWERS_PROPERTY_KEY, FilterOperator.EQUAL, user.getUserKey());
+        Query userRoomQuery = new Query(Room.ROOM_ENTITY_NAME).setFilter(userRoomFilter);
+        PreparedQuery userRoomResults = datastore.prepare(userRoomQuery);
+ 
+        HashMap<String, Integer> userTagMap = Tag.constructUserVectorMap();
+        List<String> userTags = null;
+        for(Entity roomEntity : userRoomResults.asIterable()){
+            userTags = new Room(roomEntity).getAllTags();
+            for(String tag : userTags) {
+                userTagMap.put(tag, userTagMap.get(tag) + 1);
+            }
+        }
+
+        return userTagMap;
+    }
+ 
+}
--- a/capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
+++ b/capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
@@ -34,7 +34,7 @@ public class RecommenderAlgorithm {
             otherUserVectorMap.put(otherUser, otherUserTagMap);
         }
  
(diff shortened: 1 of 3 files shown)


capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,59 @@
...
+        User otherUser = null;
+        HashMap<String, Integer> otherUserTagMap = null;
+        HashMap<User, HashMap<String,Integer>> otherUserVectorMap = new HashMap<User, HashMap<String,Integer>>();
+        for(Entity otherUserEntity : otherUserResults.asIterable()) {
+            otherUser = new User(otherUserEntity);
+            otherUserTagMap = RecommenderAlgorithm.populateUserTagMap(otherUser);
+            otherUserVectorMap.put(otherUser, otherUserTagMap);
+        }
+ 
+        // Algorithm is work in progress. Have yet to normalize and calculate relationship.
```
707090: This is exactly the kind of comment I would leave, but you want to make sure to start it with "TODO: " since that is a common practice that linters will be able to pick up on

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,59 @@
...
+    public static ArrayList<Room> recommendRooms(User user) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        HashMap<String, Integer> userTagMap = RecommenderAlgorithm.populateUserTagMap(user);
+        
+        Query otherUserQuery = new Query(User.USER_ENTITY_NAME);
+        PreparedQuery otherUserResults = datastore.prepare(otherUserQuery);
+ 
+        User otherUser = null;
+        HashMap<String, Integer> otherUserTagMap = null;
+        HashMap<User, HashMap<String,Integer>> otherUserVectorMap = new HashMap<User, HashMap<String,Integer>>();
```
707090, line 30: You should never need to load all users, as that will not scale and quickly become slow. Instead, the tag map vector should be assumed to be already up to date and cached on the user. The cached user vector only needs to be updated when users join or leave a room, not when a recommendation needs to be made

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
```diff
@@ -0,0 +1,149 @@
...
+    public void setUp() {
+        helper.setUp();
+        TEST_USER_SMITH = new User("1234", "Smith");
+        TEST_USER_JOHN = new User("2341", "John");
+        TEST_USER_BILL = new User("3412", "Bill");
+        TEST_USER_SAM = new User("4123", "Sam");
+        DatabaseService.save(TEST_USER_SMITH.getUserEntity());
+        DatabaseService.save(TEST_USER_JOHN.getUserEntity());
+        DatabaseService.save(TEST_USER_BILL.getUserEntity());
+        DatabaseService.save(TEST_USER_SAM.getUserEntity());
```
707090, line 58: Unit tests should not interact with the database at all. (In order to stay "unit"). They should be as small in scope as possible and only test a single method. If knowing that something was "saved" to the database is a critical part of what needs to be tested, a mock object should be used.

Sometimes writing such a mock object and all of the setup and expectations that requires is complicated and annoying, which can be informative about how you design your methods in the first place. Each method should be easy to test, so if it can avoid talking to the database at all, that would be ideal. (You can have separate out the logic part of the method and test that, then have an outer function that calls that along with doing whatever database saving needs to be done)

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
```diff
@@ -0,0 +1,149 @@
...
+    public void populateUserTagMapOneTagOneClassroomTest() { 
+        TEST_ROOM_RUN.addTag(Tag.TAG_FITNESS);
+        TEST_ROOM_RUN.addFollower(TEST_USER_SMITH);
+        DatabaseService.save(TEST_ROOM_RUN.getRoomEntity());
+
+        HashMap<String, Integer> userTagMap = RecommenderAlgorithm.populateUserTagMap(TEST_USER_SMITH);
+        for(String tag : TEST_CATEGORY_TAGS){
+            if(tag.equals(Tag.TAG_FITNESS)){
+                Assert.assertEquals((int) userTagMap.get(tag), 1);
+                continue;
```
707090: `else` preferred to `continue` + fallthrough since it is easier to track, and harder to accidentally make a bug out of (its pretty easy in a big nested if structure to think its OK to delete a `continue`, only to not realize that certain fallthroughs were critical). As a rule of thumb, `continue` is an antipattern (though not always conveniently avoidable)

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
```diff
@@ -0,0 +1,149 @@
...
+            }
+            Assert.assertEquals((int) userTagMap.get(tag), 0);
+        }
+    }
+
+    @Test
+    public void populateUserTagMapMultipleTagOneClassroomTest() {
```
707090: Are these still "Classroom"s?

707090 (changes requested):

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
AndrewDPham: Sounds good.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
AndrewDPham: Got it. I believe in a future PR, Kyle used embedded entities but we still need to cache the vector map when the user follows a room still. I think for now, the best approach is to leave this here since Kyle changed most of the foundation of the code and change his PR to cache the vector map when the user follows a room.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
AndrewDPham: Got it.

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
AndrewDPham: Got it. My original thinking was that since these methods pulls entities from Datastore, I should mimic that behavior. It makes more sense to use mock objects though. Thanks for the tip!

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
AndrewDPham: This makes more sense. I think I got to used to continue that I use it since it's "faster" in some cases since it doesn't finish the loop. Thanks for the tip!

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
AndrewDPham: Sorry, that is a mistake. Will change.

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
707090: Its actually not faster since the compiler breaks both the "if" and the "else" part into control flow blocks the same as it would break the "up to the continue" and "after the continue" part into control flow blocks. In any case, if it is faster its on the scale of nanoseconds. Its nearly always (except in extreme performance critical scenarios) best to optimize for easiest to read and modify without making bugs since the compiler is pretty smart and will make better optimizations automatically than you can manually.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.Entity;
+import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.List;
+import java.lang.Iterable;
+ 
+public class RecommenderAlgorithm {
+ 
+    public static ArrayList<Room> recommendRooms(User user) {
```
mattlgy, line 21: Recommend adding a comment that explains (briefly) how this works, or link to your design doc.

mattlgy (changes requested):

AndrewDPham:
Closing to create new PR.

PR #24, opened 2020-08-25
Algorithm can now return a list of recommended rooms

kyleasmith1:
Added EmbeddedEntity functionality and created methods that allow the algorithm to run smoothly.

### DIFF
--- a/capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
+++ b/capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
@@ -2,6 +2,8 @@ package com.google.sps.data;
  
 import com.google.sps.data.Room;
 import com.google.sps.data.User;
+import com.google.sps.data.Tag;
+import com.google.sps.service.DatabaseService;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.PreparedQuery;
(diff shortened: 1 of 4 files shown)


capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -11,49 +13,158 @@
...
-        return userTagMap;
+        for (Entity entity : results.asIterable()) {
+            try{
+                RecommenderAlgorithm.normalizeEmbeddedEntity(DatabaseService.getUser(entity.getKey()));
+            } catch (EntityNotFoundException e) {
+                System.out.println("Entity not found!");
+            }
+        }
     }
+
```
707090: As a general rule methods should go in the order of dependencies. So since `normalizeEmbeddedEntity` is only used by this (I think), it should go after this method. If you think of it like reading a book, I am going top to bottom, and when I get to this method Im thinking "what? what is this for?" if I dont have the context for what kind of what it will be used.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -11,49 +13,158 @@
...
+                for (String tag : room.getAllTags()) {
+                    double frequency = (Double) entity.getProperty(tag);
+                    entity.setProperty(tag, frequency + 1.0);
+                }
             }
         }
+        user.setEmbeddedTags(entity);
+    }
+
+    public static void normalizeEmbeddedEntity(User user) {
```
707090: Should this be public? As the interface stands, I this is only used by this class

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -11,49 +13,158 @@
...
-    public static ArrayList<Room> recommendRooms(User user) {
+
+    // TODO: Will add the proportions in the next PR when I know exact percentages
+
+    public static List<Key> recommendRooms(User user, Integer amount) {
+        RecommenderAlgorithm recommender = null;
+        recommender.normalizeAllEmbeddedEntities();
+
+        Map<Key, Double> distances = recommender.getDistanceMaps(user, user.getEmbeddedTags());
+        Map sorted = sortByValues(distances);
```
707090: Upcasting is dangerous here, because you actually care a lot about the specific kind of map it is. Since you are relying on the order of the map for this function to work, it is not enough for this to just be the Map type. I may see this and be like "oh I can refactor this, I have another way to get a map of all the distances, I can just delete replace this", but actually, you need sorted distances, so this should explicitly be of the TreeMap type

707090: Although... in reality, you dont need a map at all. Since all you are getting is the key, this should just return a list of Keys

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -11,49 +13,158 @@
...
-        HashMap<User, HashMap<String,Integer>> otherUserVectorMap = new HashMap<User, HashMap<String,Integer>>();
-        for(Entity otherUserEntity : otherUserResults.asIterable()) {
-            otherUser = new User(otherUserEntity);
-            otherUserTagMap = RecommenderAlgorithm.populateUserTagMap(otherUser);
-            otherUserVectorMap.put(otherUser, otherUserTagMap);
+        Query query = new Query(Room.ROOM_ENTITY_NAME);
+        PreparedQuery results = datastore.prepare(query);
+
+        if (results.countEntities() == 0) {
+            return new ArrayList<>();
```
707090: Iterating over all entities will become too slow very quickly. This should use filters

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -11,49 +13,158 @@
...
- 
-        HashMap<String, Integer> userTagMap = Tag.constructUserVectorMap();
-        List<String> userTags = null;
-        for(Entity roomEntity : userRoomResults.asIterable()){
-            userTags = new Room(roomEntity).getAllTags();
-            for(String tag : userTags) {
-                userTagMap.put(tag, userTagMap.get(tag) + 1);
+        Query query = new Query(Room.ROOM_ENTITY_NAME);
+        PreparedQuery results = datastore.prepare(query);
+        EmbeddedEntity entity = user.getEmbeddedTags();
```
707090: Why was the implementation which used a filter removed? Querying all rooms will not scale, and a filter will use indexes to make getting the rooms you are looking for quick.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -11,49 +13,158 @@
...
+            Double distance = RecommenderAlgorithm.distanceBetween(entity, (EmbeddedEntity) result.getProperty(User.TAGS_PROPERTY_KEY));
+            userDistancePairs.put(result.getKey(), distance);
+        }
+        return userDistancePairs;
+    }
+
+    public static Double distanceBetween(EmbeddedEntity userTagMap, EmbeddedEntity otherTagMap) {
+        double sum = 0;
+        for (String tag : Tag.CATEGORY_TAGS) {
+            double value = (Double) userTagMap.getProperty(tag) - (Double) otherTagMap.getProperty(tag);
```
707090: a variable name like delta helps indicate what the variable means

707090 (changes requested):
As a general rule, any query which queries for ALL of any entity will not scale. Always use filters

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
kyleasmith1: good point, will change the return type

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
kyleasmith1: no it should be private you are right

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
kyleasmith1: I couldn't figure out a filter that would work for this part

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
kyleasmith1: gotcha... I didn't even notice that Andrew had gotten the filter to work. Didn't think that filter would do anything because I thought it would check if the rooms followers list was equal to the user but I guess it checks if the user is in the rooms followers list

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
kyleasmith1: gotcha

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
kyleasmith1: only reason I used TreeMaps is so that I could sort the values. If I just returned a list of keys I could do the sorting in the method I guess?

mattlgy (approved):
Looks good, after addressing Scott's comment.

AndrewDPham:
Closing to create new PR.

PR #27, opened 2020-08-25
Client tag

AndrewDPham:
Add functionality for users to add tags to rooms on creation.
Screenshot 2020-08-25 at 1 55 13 PM

### DIFF
--- a/capstone/src/main/webapp/dashboard-loader.js
+++ b/capstone/src/main/webapp/dashboard-loader.js
@@ -2,7 +2,26 @@ document.getElementById("signout").prepend(dynamicButton);
 window.addEventListener('authorized', getRooms);
 
 function createRoom(title, description) {
-    var roomData = JSON.stringify({"title": title, "description": description});
+    var checkboxes = document.getElementsByClassName("form-check-input");
+    var tagsData = [];
+    for(checkbox of checkboxes){
+        if(checkbox.checked === true){
+            console.log("Added");
+            tagsData.push(checkbox.value);
+        }
+    }
+    if(tagsData.length < 1 || tagsData.length > 3){
+        console.log("Diff number of tags");
+        return;
+    }
+
+    $(document).ready(function(){
+        $("#modalBtn").click(function(){
+            $("#exampleModalCenter").modal("hide");
+        });
+    });
+
+    var roomData = JSON.stringify({"title": title, "description": description, "tags": tagsData});
     fetch("/dashboard", {method: "POST", headers: new Headers({ID_TOKEN}), body: roomData}).then((resp) => {
         if (resp.ok){
             getRooms();
--- a/capstone/src/main/webapp/dashboard-loader.js
+++ b/capstone/src/main/webapp/dashboard-loader.js
@@ -6,20 +6,15 @@ function createRoom(title, description) {
     var tagsData = [];
     for(checkbox of checkboxes){
         if(checkbox.checked === true){
-            console.log("Added");
             tagsData.push(checkbox.value);
         }
     }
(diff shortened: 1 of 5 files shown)


capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -2,7 +2,21 @@ document.getElementById("signout").prepend(dynamicButton);
 window.addEventListener('authorized', getRooms);
 
 function createRoom(title, description) {
-    var roomData = JSON.stringify({"title": title, "description": description});
+    var checkboxes = document.getElementsByClassName("form-check-input");
```
707090, line 5: I would give the checkboxes by a semantic class instead of relying on a bootstrap class. Because supposing you ever decided to change to not use Bootstrap, that would break actual application logic.

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -41,17 +55,59 @@ function createRoomDivElement(room) {
                     <h5 class="card-title" id="room-title"></h5>
                     <a class="card-text small-text" href="#">Kyle Smith</a>
                     <div class="card-text small-text">Followers: Infinite</div>
-                    <div class="card-text small-text">Tag(s): </div>
+                    <div class="card-text small-text" id="room-tags"></div>
```
707090, line 58: If this is something that is repeated multiple times, using ID would create invalid HTML (Applies everywhere). Instead use a class name and look it up that way

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -41,17 +55,59 @@ function createRoomDivElement(room) {
...
+    return doc.body;
+}
+
+function closeModal(){
+    document.getElementById("title").value = "";
+    document.getElementById("description").value = "";
+    
+    var checkboxes = document.getElementsByClassName("form-check-input");
+    for(checkbox of checkboxes){
+        if(checkbox.checked === true){
```
707090, line 91: Do you need this check? You can always just set it to false even if it already was.

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -41,17 +55,59 @@ function createRoomDivElement(room) {
...
     doc.getElementById("room-link").innerText = "View";
     return doc.body;
+}
+
+function createBadgeDivElement(tag){
+    let domparser = new DOMParser();
+    let doc = domparser.parseFromString(`
+        <span class="badge badge-pill badge-secondary" id="tag"></span>
+    `, "text/html");
+    doc.getElementById("tag").innerText = tag;
```
707090, line 81: Do not use ID for tag which might appear more than once

707090 (changes requested):

capstone/src/main/webapp/dashboard-loader.js
mattlgy, line 5: You could do something like:
`document.querySelectorAll('input[name=tag-checkboxes]:checked')`
to get just the checked ones. Then you would not have to check `checked` below.

Could could even get real fancy and do something like:
```javascript
const tagsData =
    document.querySelectorAll('input[name=tag-checkboxes]:checked')
    .map((checkbox) => checkbox.value);
```

The above selector would require adding a name attribute to your checkboxes below, or you could select by class, `.form-check-input:checked`, but I think added a semantically  meaningful name would be better.

mattlgy (changes requested):
Prefer to use `let` and `const` in JS rather than var.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

PR #29, opened 2020-08-31
Added video visualization for lessons

kyleasmith1:


capstone/src/main/webapp/dashboard.html
```diff
@@ -67,7 +67,7 @@ <h5 class="modal-title" id="exampleModalCenterTitle">Pick A Room Title</h5>
     <div class="container-fluid">
         <div id="room-container" class="row row-cols-1 row-cols-sm-4"></div>
     </div>
- 
+
```
kyleasmith1, line 70: Cloud shell was being weird about this white space and I couldn't figure out how not to commit it sorry.

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -42,23 +43,33 @@ function getSearchResults(type='') {
     }); 
 }
 
+function followerCount(room) {
+    var count = 0
+    if (room.entity.propertyMap.followers == null) {
+        return "Follower: " + count;
+    }
```
707090, line 50: Inline count

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -42,23 +43,33 @@ function getSearchResults(type='') {
...
+    return "Follower: " + room.entity.propertyMap.followers.length;
```
707090, line 51: You actually dont need to inline or have the if check at all because you can use the nullish operator to check for null:

return "Follower: " + room.entity.propertyMap.followers.length ?? 0;

capstone/src/main/webapp/script.js
```diff
@@ -24,22 +24,23 @@ function onSignIn(googleUser) {
...
+    }
+    return "Description: " + capitalizeFLetter(lesson.entity.propertyMap.description);
+}
+
+function capitalizeFLetter(string) { 
+    if (string == "") {
+        return "Missing Input";
+    }
+    return string.replace(/^./, string[0].toUpperCase()); 
+}
```
707090, line 46: F letter? This doesnt capitallize Fs, it just capitalizes the first letter. Unless F is short for first? In which case it is difficult to tell that meaning from other parts of the code.

capstone/src/main/webapp/script.js
```diff
@@ -24,22 +24,23 @@ function onSignIn(googleUser) {
...
+function getDescription(object) {
+    if (object.entity.propertyMap.description == "") {
+        return "";
+    }
+    return "Description: " + capitalizeFLetter(lesson.entity.propertyMap.description);
+}
+
+function capitalizeFLetter(string) { 
+    if (string == "") {
+        return "Missing Input";
```
707090, line 43: This would be quite confusing to get back from capitalization of an empty string. Wouldnt it make sense to just return an empty string and just treat that as though the first non-letter can just be treated as capital?

capstone/src/main/webapp/lesson-loader.js
```diff
@@ -70,55 +70,80 @@ function createContent(title, description, content, urls) {
...
 function getRoomId() {
     return new URL(window.location.href).searchParams.get("room_id");
 }
 
 function getJoinStatus() {
-    var action = new URL(window.location.href).searchParams.get("action")
+    var action = new URL(window.location.href).searchParams.get("action");
     if (action == null) {
         return "unjoin";
     }
```
707090, line 89: This is another instance where you can use the nullish operator to check for null and apply a default:

return new URL(window.location.href).searchParams.get("action") ?? "unjoin";

capstone/src/main/webapp/lesson.html
```diff
@@ -121,15 +122,15 @@ <h5 class="modal-title" id="ModalCenterTitle">Content</h5>
                 </div>
                 <div class="modal-body">
                         <p> Content Information: </p>
-                        <input type="text" id="title" placeholder="Title" required>
-                        <input type="text" id="description" placeholder="Description" required>  
+                        <input type="text" id="content-title" placeholder="Title" required>
+                        <input type="text" id="content-description" placeholder="Description" required>  
                         <input type="text" id="content" placeholder="Content" required>
-                        <input type="text" id="urls" placeholder="URL(s)" required>
+                        <input type="text" id="content-urls" placeholder="URL(s)" required>
```
707090, line 128: These should work fine as classes probably? At least for title and description, it seems like setting a nice generic class would allow you to keep the targetting code the same as well as potentially use common CSS across all of them

707090 (changes requested):

PR #30, opened 2020-08-31
Add content feature to web application

kyleasmith1:


capstone/src/main/webapp/lesson-loader.js
```diff
@@ -90,12 +90,22 @@ function getJoinStatus() {
...
+function createHyperLink(url) {
+    let domparser = new DOMParser();
+    let doc = domparser.parseFromString(`
+        <a id="object-url"></a>
+    `, "text/html");
+    doc.getElementById("object-url").href = url;
+    doc.getElementById("object-url").innerHTML = url;
+    return doc.body;
+}
+
```
707090, line 102: Having the ID set means that if this is called multiple times, it would create invalid HTML (since only 1 thing is allowed to have an ID per page). This applies to all instances on this page, but in this case its simpler as just

let anchor = document.createElement("a");
anchor.href = url;
anchor.innerHTML = url;
return anchor;
You want to be careful to always re-think through anytime you copy and paste code or an approach.

Also I would call this method anchor instead of hyperlink (which I think is actually 1 word).

capstone/src/main/webapp/lesson-loader.js
```diff
@@ -146,4 +156,19 @@ function createVideoDivElement(lesson) {
...
+function createContentDivElement(lesson) {
+    let domparser = new DOMParser();
+    let doc = domparser.parseFromString(`
+        <div id="modal-content"></div>
+        <div class="small-spacing-bottom"></div>
+        <div id="modal-title"><span>Link(s)</span></div>
+        <p><div id="modal-urls"></div></p>
+    `, "text/html");
+    doc.getElementById("modal-content").innerHTML = capitalizeFLetter(lesson.entity.propertyMap.content);
```
mattlgy, line 169: I think this should be `.innerText = ...`.

capstone/src/main/webapp/lesson-loader.js
```diff
@@ -146,4 +156,19 @@ function createVideoDivElement(lesson) {
     `, "text/html");
     doc.getElementById("player").src = videoPath + new URL(lesson.entity.propertyMap.url).searchParams.get("v");
     return doc.body;
+}
+
+function createContentDivElement(lesson) {
+    let domparser = new DOMParser();
+    let doc = domparser.parseFromString(`
+        <div id="modal-content"></div>
```
mattlgy, line 164: As mentioned, IDs should not be duplicated.

mattlgy (approved):

707090 (changes requested):

PR #31, opened 2020-08-31
Added form feature to lessons

kyleasmith1:


capstone/src/main/webapp/lesson-loader.js
```diff
@@ -171,4 +171,17 @@ function createContentDivElement(lesson) {
...
     }
     return doc.body;
+}
+
+function createFormDivElement(lesson) {
+    let domparser = new DOMParser();
+    let doc = domparser.parseFromString(`
+        <div id="modal-form-description"></div>
+        <div id="modal-title"><span>Form</span></div>
+        <a id="modal-form-url"></a
```
mattlgy, line 181: As Scott mentioned, avoid using ids for elements that will be duplicated.

mattlgy (approved):

PR #32, opened 2020-08-31
Tag update

AndrewDPham:
Changes to tag class to make it of object "Tag". Add more properties to Users


capstone/src/main/java/com/google/sps/data/User.java
```diff
@@ -32,6 +38,14 @@ public String getNickname() {
         return (String) this.delegate.getProperty(User.NICKNAME_PROPERTY_KEY);
     }
 
+    public void setEmbeddedTags(EmbeddedEntity entity) {
```
707090: As discussed, I would name this something other than "EmbeddedTags" which indicates what it is

707090 (approved):

capstone/src/main/java/com/google/sps/data/User.java
AndrewDPham: Got it.

PR #33, opened 2020-09-01
Add delete functionality (for Lessons and Rooms)

kyleasmith1:


capstone/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
```diff
@@ -0,0 +1,48 @@
...
+
+    if (kind.equals(Room.ROOM_ENTITY_NAME)) {
+        try {
+            for (Key lessonKey : DatabaseService.getRoom(key).getAllLessons()) {
+                DatabaseService.delete(lessonKey);
+            }
+        } catch (EntityNotFoundException e) {
+            System.out.println("Lesson entity does not exist: " + e);
+        }
+    }
```
707090, line 43: This kind of logic doesnt belong in the servlet, but in the database service. What if some other part of the code called the existing database service's method on a room, but didnt do this? Instead the database service should have delete methods for each kind, which understands the data model and does appropriate cleanup like this on a per-entity basis. (For example when users are deleted, they should also be removed from all rooms followers)

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
```diff
@@ -64,15 +65,44 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    // Checks to see if the lesson still exists in datastore
+    public Boolean checkLessonStatus(Key key) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Query query = new Query(Lesson.LESSON_ENTITY_NAME).addFilter(Entity.KEY_RESERVED_PROPERTY, FilterOperator.EQUAL, key);
+        Entity result = datastore.prepare(query).asSingleEntity();
+
+        if (result == null) {
+            return false;
+        }
+        return true;
```
707090, line 97: Anytime you have:
```java
if (condition) {
    return false;
} else {
    return true;
}
```
you can simplify to `return !condition;`

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
```diff
@@ -64,15 +65,44 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
         }
         catch (EntityNotFoundException e) {
             System.err.println("Lesson entities don't exist.");
         }
+        removeDeletedLessons(room, nullKeys);
         return lessons;
     }
+
+    // Checks to see if the lesson still exists in datastore
+    public Boolean checkLessonStatus(Key key) {
```
707090, line 89: You should basically never use the Boolean type since autoboxing will handle turning a primitive `boolean` into a Boolean when it needs to be cast to an object.

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
```diff
@@ -64,15 +65,44 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    }
+
+    public void removeDeletedLessons(Room room, List<Key> keys) {
+        List<Key> lessons = room.getAllLessons();
+        for (Key key : keys) {
+            lessons.remove(key);
+        }
+        room.getRoomEntity().setProperty(Room.LESSONS_PROPERTY_KEY, lessons);
+        DatabaseService.save(room.getRoomEntity());
+    }
```
707090, line 107: This kind of cleanup (removing the lesson key from the parent room) should be done ahead of time when the lesson is removed (part of the "appropriate cleanup" relevant to lessons mentioned above), not upon requesting lessons. (This means this file should have basically no diff other than moving what was already existing for the getLessons to DatabaseService)

capstone/src/main/java/com/google/sps/servlets/LessonHandlerServlet.java
```diff
@@ -64,15 +65,44 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+                    nullKeys.add(key);
+                }
             }
         }
         catch (EntityNotFoundException e) {
             System.err.println("Lesson entities don't exist.");
         }
+        removeDeletedLessons(room, nullKeys);
         return lessons;
     }
```
707090, line 86: This method should probably live in the DatabaseService, since it could be re-used by any other part of the code which needs a full list of Lessons from a room.

capstone/src/main/webapp/dashboard.html
```diff
@@ -63,6 +63,23 @@ <h5 class="modal-title" id="exampleModalCenterTitle">Pick A Room Title</h5>
...
+        <div class="modal-dialog modal-dialog-centered" role="document">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h5 class="modal-title">Are You Sure You Want to Delete This Room?</h5>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                    <span aria-hidden="true">&times;</span>
+                    </button>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-outline-danger" id ="room-delete-button" data-dismiss="modal">Yes</button>
```
707090, line 77: whitespace `id ="...`

capstone/src/main/webapp/dashboard-loader.js
```diff
@@ -62,17 +62,19 @@ function createRoomDivElement(room) {
...
     doc.getElementById("room-title").innerText = capitalizeFLetter(room.entity.propertyMap.title);
     doc.getElementById("room-description").innerText = getDescription(room);
     doc.getElementById("room-followers").innerText = followerCount(room);
     doc.getElementById("room-link").addEventListener("click", function() {
         window.location.href = "lesson.html?room_id=" + room.entity.key.id;
     });
-    doc.getElementById("room-link").innerText = "View";
+    document.getElementById("room-delete-button").addEventListener("click", function() {
+        deleteFromDatastore(room);
+    });
```
707090, line 78: I would just have this be an on-click directly on the element (applies to all of them)

capstone/src/main/webapp/dashboard.html
```diff
@@ -63,6 +63,23 @@ <h5 class="modal-title" id="exampleModalCenterTitle">Pick A Room Title</h5>
             </div>
         </div>
     </div>
+
+    <div class="modal fade" id="ModalCenterRoomDelete" tabindex="-1" role="dialog" aria-labelledby="modal-type" aria-hidden="true">
+        <div class="modal-dialog modal-dialog-centered" role="document">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h5 class="modal-title">Are You Sure You Want to Delete This Room?</h5>
```
707090, line 71: (optional) The Title Case Seems Unusual For A Normal Message?

capstone/src/main/webapp/script.js
```diff
@@ -43,4 +43,15 @@ function capitalizeFLetter(string) {
...
     return string.replace(/^./, string[0].toUpperCase()); 
+}
+
+function deleteFromDatastore(object) {
+    var keyData = JSON.stringify({"kind": object.entity.key.kind, "id": object.entity.key.id});
+    fetch("/delete", {method: "POST", headers: new Headers({ID_TOKEN}), body: keyData}).then((resp) => {
+        if (resp.ok){
+            location.reload();
+        } else {
+            alert("Error has occured");
```
707090, line 54: Its important that errors are always as informative as possible. If someone files a bug with your software which say "I got an alert saying 'Error has occurred'" you dont really have anything to go off of. Its also confusing as a user when there is no detail given, because how will I know what to do now?

707090 (changes requested):

PR #34, opened 2020-09-01
Vector map update

AndrewDPham:
Create CachedInterestVector class to house static helper methods to update interest vector of user.


capstone/src/main/java/com/google/sps/data/CachedInterestVector.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public static EmbeddedEntity hashMapToEmbeddedEntity(HashMap<String, Double> vectorHashMap){
+        EmbeddedEntity embeddedVectorMap = new EmbeddedEntity();
+        for(String tag : vectorHashMap.keySet()) {
+            embeddedVectorMap.setProperty(tag, vectorHashMap.get(tag));
+        }
+        return embeddedVectorMap;
+    }
+
+    public static void denormalizeVectorHashMap(HashMap<String, Double> vectorHashMap, Double magnitude) {
```
707090, line 33: This would be better to return a completely separate HashMap instead of modifying the existing one in place. A couple benefits of this are 1. You can improve typing by returning a `HashMap<String, Integer>` (which will also change the signature of some of the other functions) since that is more representative of what a denormed map should be. 2. If somehow a bug cropped up and this threw an exception, then the map would be in a mutant half-normalized state. This is probably pretty unlikely in this particular function, but the practice of modifying things in place and using that as the "return" value even though its just a parameter is not best practice because it can be hard to spot this problem.

capstone/src/main/java/com/google/sps/data/User.java
```diff
@@ -46,6 +46,14 @@ public EmbeddedEntity getCachedInterestVector() {
         return (EmbeddedEntity) this.delegate.getProperty(User.TAGS_PROPERTY_KEY);
     }
 
+    public void setMagnitude(Double magnitude) {
```
707090, line 49: getMagnitude should probably have a name which associates it with the cached interest vector so it doesnt seem like it is a normalized property of user. (at least include the name cached, and probably a name which associates it with the vector as well)

capstone/src/main/java/com/google/sps/data/CachedInterestVector.java
```diff
@@ -0,0 +1,79 @@
...
+            } else {
+                vectorHashMap.put(tag.getTag(), vectorHashMap.get(tag.getTag()) + 1);
+            }
+        }
+    }
+
+    public static void removeTagFromDenormalizedVectorHashMap(HashMap<String, Double> vectorHashMap, List<Tag> tags) {
+        for(Tag tag : tags) {
+            if (vectorHashMap.get(tag.getTag()) == null) {
+                continue;
```
707090, line 55: Worth adding a comment about our discussion on this line for future readers

capstone/src/main/java/com/google/sps/data/CachedInterestVector.java
```diff
@@ -0,0 +1,79 @@
...
+            sum += Math.pow(value, 2.0);
+        }
+        return Math.sqrt(sum);
+    }
+ 
+    public static void renormalizeVectorHashMap(HashMap<String, Double> vectorHashMap, Double magnitude) {
+        for(String tag : vectorHashMap.keySet()) {
+            vectorHashMap.put(tag, ((vectorHashMap.get(tag))/magnitude));
+        }
+    }
```
707090, line 78: Order this function next to the denorm one. Also name should not be "re-" since the method has no specific requirements about where this map came from. I can hand-create a map which was never normalized in the first place and it still works fine.

707090, line 78: Magnitude should not have to be passed in separately, since it is derivable information and can be calculated with a method call right in the function. (Because it would be nonsense for me to call this with `vectorHashMap: { "edu": 4, "cooking": 4 }, magnitude: 2.0`)

capstone/src/test/java/com/google/sps/CachedInterestVectorTest.java
```diff
@@ -0,0 +1,148 @@
...
+        EmbeddedEntity embeddedVectorMap = new EmbeddedEntity();
+        embeddedVectorMap.setProperty(Tag.EDUCATION.getTag(), (2.0/magnitude));
+        embeddedVectorMap.setProperty(Tag.COOKING.getTag(), (2.0/magnitude));
+        embeddedVectorMap.setProperty(Tag.FITNESS.getTag(), (1.0/magnitude));
+
+        HashMap<String, Double> vectorHashMap = CachedInterestVector.embeddedEntityToHashMap(embeddedVectorMap);
+        Assert.assertEquals(vectorHashMap.get(Tag.EDUCATION.getTag()), (2.0/magnitude), epsilon);
+        Assert.assertEquals(vectorHashMap.get(Tag.COOKING.getTag()), (2.0/magnitude), epsilon);
+        Assert.assertEquals(vectorHashMap.get(Tag.FITNESS.getTag()), (1.0/magnitude), epsilon);
+    }
```
707090, line 55: Add to the test (or a separate test if you want) what happens when you explicitly put null as the value for a property. What about zero? What about a negative number? Since this is simply a conversion method to swap entities and hash maps, Im not necessarily saying error detection needs to happen here (even if negative numbers are invalid, then it might not throw and might just have a negative in the hashmap since validation is not the goal of this method: thats fine, but it should be demonstrated in the test)

Similar cases for the reverse conversion method test below

capstone/src/test/java/com/google/sps/CachedInterestVectorTest.java
```diff
@@ -0,0 +1,148 @@
...
+        tagsToRemove.add(Tag.EDUCATION);
+        tagsToRemove.add(Tag.COOKING);
+        tagsToRemove.add(Tag.FITNESS);
+
+        CachedInterestVector.removeTagFromDenormalizedVectorHashMap(vectorHashMap, tagsToRemove);
+
+        Assert.assertEquals(vectorHashMap.get(Tag.EDUCATION.getTag()), 1.0, epsilon);
+        Assert.assertEquals(vectorHashMap.get(Tag.COOKING.getTag()), 1.0, epsilon);
+        Assert.assertNull(vectorHashMap.get(Tag.FITNESS.getTag()));
+    }
```
707090, line 122: Test (this or another) should cover what happens when the tag was already not present, but is asked to be removed anyway

707090 (changes requested):

capstone/src/main/java/com/google/sps/data/CachedInterestVector.java
```diff
@@ -0,0 +1,79 @@
...
+ 
+    public static Double magnitude(HashMap<String, Double> vectorHashMap) {
+        double sum = 0;
+        for (String tag : vectorHashMap.keySet()) {
+            double value = vectorHashMap.get(tag);
+            sum += Math.pow(value, 2.0);
+        }
+        return Math.sqrt(sum);
+    }
```
707090, line 72: What if the sum is still zero because the map is empty? This will throw an exception and probably crash things. Adjust code then add a test to cover it

PR #35, opened 2020-09-01
Update cached vector

AndrewDPham:
Add update methods to update interest vector of users when they follow/unfollow rooms.


capstone/src/main/java/com/google/sps/servlets/FollowServlet.java
```diff
@@ -62,6 +63,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
             room.removeFollower(user);
         }
 
+        DatabaseService.save(user.getUserEntity());
         DatabaseService.save(room.getRoomEntity());
```
707090, line 67: As a note: this right here is totally correct given the context but a little unfortunate. Because calling room.addFollower actually updates the user itself, you have to remember to save this. I might in another context not realize that adding a follower actually modifies the user as well due to caching and not realize I need to save the user. This can be a source of bugs and problematic conceptually, and is a huge benefit of using an ORM of some kind (thats what its named for SQL dbs, idk what the NoSQL equivalent is called), which automatically does the Object <-> Database mapping and save calls and everything and lets you just treat the documents/rows in a database as though they are just your model objects.

capstone/src/test/java/com/google/sps/CachedInterestVectorTest.java
```diff
@@ -145,4 +145,47 @@ public void renormalizeVectorHashMapTest() {
...
+    
+    @Test
+    public void removeRoomUpdateCachedInterestVectorTest() {
+        User user = new User("123" , "John");
+        Room room = new Room(user, "Piano", "Piano Lessons");
+        room.addTag(Tag.EDUCATION);
+        room.addTag(Tag.MUSIC);
+        room.addTag(Tag.ART);
+        EmbeddedEntity embeddedVectorMap = new EmbeddedEntity();
+        Double magnitude = Math.sqrt(15.0);
```
707090, line 175: 3^2 + 2^2 + 1^2 = 14. If this passes, perhaps your epsilon is a little to large? To avoid this kind of thing happening, I would re-use the same setup you use in your actual code, since each test can assume the other methods are propertly tested. So I would do something like:
```java
HashMap<String, Double> preexisting_tags = new HashMap<>();
preexisting_tags.put(...);
user.setCachedInterestVector(hashMapToEmbeddedEntity(preexisting_tags));
user.setMagnitude(magnitude(preexisting_tags));
```

707090 (approved):

capstone/src/test/java/com/google/sps/CachedInterestVectorTest.java
AndrewDPham: It passes since I am passing in the magnitude too (which you commented on another PR that the magnitude shouldn't be passed in) so it should be 14.0. Got it.

capstone/src/main/java/com/google/sps/servlets/FollowServlet.java
AndrewDPham: Got it. That's an interesting concept, I will look more into it.

PR #37, opened 2020-09-02
Recommender Algo Update

AndrewDPham:
Add RecommenderAlgorithm Class to recommender rooms to users.


capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
```diff
@@ -0,0 +1,422 @@
...
+        EmbeddedEntity johnTagMap = TEST_USER_JOHN.getCachedInterestVector();
+        EmbeddedEntity samTagMap = TEST_USER_SAM.getCachedInterestVector();
+
+        Double smithJohnDistance = RecommenderAlgorithm.distanceBetween(smithTagMap, johnTagMap);
+        Double smithSamDistance = RecommenderAlgorithm.distanceBetween(smithTagMap, samTagMap);
+
+        Assert.assertEquals(smithJohnDistance, 0.33820395745152554, epsilon);
+    }
+
+    /* ------------------------------------------------------------ */
```
AndrewDPham, line 223: All the test above do not need to use DatabaseService.save(). For the tests below, we really should be testing for findSimilarRooms() and change the function implementation so that it doesn't talk to the database. Because of time, I decided to leave the test as is and in a future branch (we all know this won't happen xD) we can remove the DatabaseService.save() and further separate application and intregration logic.

AndrewDPham: Also most of these tests are credited to Kyle. I decided to copy and paste from PR#25 instead of merging since it seemed easier so it doesn't have his commits. All I changed was the tag name and remove the normalize embedded entity.

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
```diff
@@ -0,0 +1,422 @@
...
+ 
+    @After
+    public void tearDown() {
+        helper.tearDown();
+    }
+
+    @Test
+    public void calculateDistanceSameUserTest() { 
+        TEST_ROOM_RUN.addTag(Tag.FITNESS);
+        TEST_ROOM_RUN.addFollower(TEST_USER_SMITH);
```
AndrewDPham, line 87: @707090 It was a really good call to update the cached interest vector when we add the follower to the room since here, all we need to call is addFollower() and check the cached interest vector for the tests.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+import java.lang.Iterable;
+import java.util.Collections;
+import java.util.Comparator;
+import java.lang.Math;
+import java.lang.Iterable;
+ 
+public class RecommenderAlgorithm {
+
+    public static List<Key> recommendRooms(User user, Integer amount) {
+        RecommenderAlgorithm recommender = null;
```
707090, line 32: This should not exist, since you never use an instance of recommender, and its only usage is to do static calls on an instance variable (which is bad practice anyway). Instead replace where you have `recommender.findSimilarRooms` with `RecommenderAlgorithm.findSimilarRooms`

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+            return rooms;
+        }
+
+        return rooms.subList(0, amount);
+    }
+
+    private static TreeMap<Key, Double> getDistanceMaps(User user, Iterable<Entity> allOtherUsers) {
+        TreeMap<Key, Double> userDistancePairs = new TreeMap<>();
+        for (Entity otherUser : allOtherUsers) {
+            Double distance = RecommenderAlgorithm.distanceBetween(user.getCachedInterestVector(), (EmbeddedEntity) otherUser.getProperty(User.TAGS_PROPERTY_KEY));
```
707090, line 58: This evades the data model and accesses a property directly. This would break if for example you decided to change the way users stored their interest vectors. Instead this method should take a list of Users and just ask them for their interest vector in whatever way they decide, then compare those.

The way I was saying to always question your code I would ask: how could someone mess up in calling this? With any `Iterable<Entity>` its very easy to mess up, not only if User changed the way it stored interest vectors, but also by just accidentally passing an Iterable<Entity> of Room or something. If the type system enforces that it is a List<User> you cant make that mistake

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+        RecommenderAlgorithm recommender = null;
+
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Filter tagFilter = new FilterPredicate(User.USER_ID_PROPERTY_KEY, FilterOperator.NOT_EQUAL, user.getId());
+        Query query = new Query(User.USER_ENTITY_NAME).setFilter(tagFilter);
+        PreparedQuery results = datastore.prepare(query);
+        Iterable<Entity> allOtherUsers = results.asIterable();
+
+        TreeMap<Key, Double> distances = RecommenderAlgorithm.getDistanceMaps(user, allOtherUsers);
+        TreeMap sorted = sortByValues(distances);
```
707090, line 41: Rather than creating a TreeMap, which is inherently already sorted by _something_, and then after the fact copying it over to another tree map but this time with a comparator for sorting how you want, you should just have the original use the comparator you want and then there is no need to copy the TreeMap

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+
+        List<Key> rooms = new ArrayList<>();
+        for (Map.Entry<Key, Double> entry : distances.entrySet()) {
+            rooms = recommender.findSimilarRooms(user, entry.getKey(), rooms);
```
707090, line 45: This is not a good interface, as it looks as though you are completely re-assigning the list to a brand new thing each time. (Further comment left on findSimilarRooms)

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        if (rooms.size() < amount) {
+            return rooms;
+        }
+
+        return rooms.subList(0, amount);
+    }
+
+    private static TreeMap<Key, Double> getDistanceMaps(User user, Iterable<Entity> allOtherUsers) {
```
707090, line 55: Why is TreeMap being used here? Is the goal to keep things sorted by the user key? This should be HashMap if it does not care about sorting. (Or really... still be TreeMap if you make the adjustment commented in recommendRooms, but this comment would apply to this method otherwise)

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+        TreeMap<Key, Double> userDistancePairs = new TreeMap<>();
+        for (Entity otherUser : allOtherUsers) {
+            Double distance = RecommenderAlgorithm.distanceBetween(user.getCachedInterestVector(), (EmbeddedEntity) otherUser.getProperty(User.TAGS_PROPERTY_KEY));
+            userDistancePairs.put(otherUser.getKey(), distance);
+        }
+        return userDistancePairs;
+    }
+
+    public static Double distanceBetween(EmbeddedEntity userTagMap, EmbeddedEntity otherTagMap) {
```
707090, line 64: You should almost never use `Double` except when you are forced to (for example you cant `extend` primitives). Other than that autoboxing will handle any primitive to Object conversions you need, and you should always use `double` everywhere

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+                continue;
+            } else if (userTagMap.getProperty(tag.getTag()) == null) {
+                delta = (Double) otherTagMap.getProperty(tag.getTag());
+            } else if (otherTagMap.getProperty(tag.getTag()) == null) {
+                delta = (Double) userTagMap.getProperty(tag.getTag());
+            } else {
+                delta = (Double) userTagMap.getProperty(tag.getTag()) - (Double) otherTagMap.getProperty(tag.getTag());
+            }
+            sum += Math.pow(delta, 2.0);
+        }
```
707090, line 78: This pattern of declaring the variable forward, and then relying on it being set by the time you get to the end of the for loop is not ideal. For example, lets say we changed the logic in the first if condition and decided we didnt need to `continue` but didnt set delta because we did some other logic to the sum directly. Then the line at the end would still be adding the square of the delta but it would be adding the delta from the last iteration of the loop.

This is also partly why I was saying `continue` is bad practice and should be avoided whenever possible.

Instead just modify the sum directly in each of the conditions and dont declare any delta varable.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+        return userDistancePairs;
+    }
+
+    public static Double distanceBetween(EmbeddedEntity userTagMap, EmbeddedEntity otherTagMap) {
+        double sum = 0;
+        double delta = 0;
+        for (Tag tag : Tag.CATEGORY_TAGS) {
+            if (userTagMap.getProperty(tag.getTag()) == null && otherTagMap.getProperty(tag.getTag()) == null) {
+                continue;
```
707090, line 69: This condition does not need to be here at all. Instead the other conditions can incorporate avoiding this condition themselves. So the next if can be `if (userTagMap.getProperty(tag.getTag()) == null && otherTagMap.getProperty(tag.getTag()) != null)`. This lets me see the complete condition that triggers this block very easily without having to have context and have read every condition in the block.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+                delta = (Double) userTagMap.getProperty(tag.getTag());
+            } else {
+                delta = (Double) userTagMap.getProperty(tag.getTag()) - (Double) otherTagMap.getProperty(tag.getTag());
+            }
+            sum += Math.pow(delta, 2.0);
+        }
+        return Math.sqrt(sum);
+    }
+
+    private static List<Key> findSimilarRooms(User user, Key key, List<Key> rooms) {
```
707090, line 82: This interface makes it look like findSimilarRooms is returning a completely new set of rooms, but its actually modifying an input variable (not recommended when avoidable) and then returning it back. Fundamentally, when I ask "what do you need to find similar rooms from this user" the list of rooms that you are already looking at is irrelevant. If I return rooms that you arent interested in, then you can just discard those yourself. Its not fundamental to finding similar rooms. This should not accept a rooms list and modify it, but create a brand new list of all relevant rooms and return it. Then if whoever calls this happens to not be interested in duplicates, they can use a Set and add the list this returns to that set. But if they are interested in duplicates, then they still have that information available.

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
```diff
@@ -0,0 +1,119 @@
...
+        if (results.countEntities() == 0) {
+            return new ArrayList<>();
+        }
+
+        for (Entity result : results.asIterable()) {
+            Room room = new Room(result);
+            if (!room.getAllFollowers().contains(user.getUserKey()) && !rooms.contains(room.getRoomKey())) {
+                rooms.add(room.getRoomKey());
+            }
+        }
```
707090, line 97: Avoid manually doing filtering when possible. In this case, you should be adding a compound filter which also incoporates a filter:
`new FilterPredicate(Room.FOLLOWERS_PROPERTY_KEY, FilterOperator.NOT_EQUAL, user.getUserKey);`

(as mentioned in the above comment, de-duping rooms returned should not go here, so that part shouldnt be necessary)

capstone/src/main/java/com/google/sps/data/RecommenderAlgorithm.java
707090, line 82: What does `key` mean? I cant tell either by looking at the signature or even by lookin through the method (except that I can figure out it is probably a user key, but arent I already passing a user?). Variable naming is very important

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
```diff
@@ -0,0 +1,422 @@
...
+        TEST_ROOM_COOK.addFollower(TEST_USER_SMITH);
+        TEST_ROOM_COOK.addFollower(TEST_USER_JOHN);
+        DatabaseService.save(TEST_ROOM_COOK.getRoomEntity());
+
+        TEST_ROOM_COOK_TWO.addTag(Tag.COOKING);
+        TEST_ROOM_COOK_TWO.addFollower(TEST_USER_BILL);
+        TEST_ROOM_COOK_TWO.addFollower(TEST_USER_SAM);
+        DatabaseService.save(TEST_ROOM_COOK_TWO.getRoomEntity());
+
+        List<Key> rooms = new ArrayList<>();
```
707090, line 416: (Applies everywhere) A good variable name for the thing you think should be the outcome is `expected` or `expected_x`

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
```diff
@@ -0,0 +1,422 @@
...
+        rooms.add(TEST_ROOM_COOK_TWO.getRoomKey());
+
+        List<Key> recommended = RecommenderAlgorithm.recommendRooms(TEST_USER_SMITH, 10);
+        Assert.assertEquals(recommended, rooms);
+    }
```
707090, line 402: Rooms get recommended even with zero tag similarity? How is distance calculated in this case since both users have manitude zero, and dividing by zero to normalize that vector would break

capstone/src/test/java/com/google/sps/RecommenderAlgorithmTest.java
```diff
@@ -0,0 +1,422 @@
...
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EmbeddedEntity;
+import com.google.appengine.api.datastore.Key;
+import com.google.sps.service.DatabaseService;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.HashMap;
+ 
+@RunWith(JUnit4.class)
+public final class RecommenderAlgorithmTest {
```
707090, line 33: None of these tests recommend the most common scenario in which there are more people and rooms than need to be recommended, and its able to reccomend the best ones. All of them end up asserting that every room in the test gets recommended, not that the best ones do. Add at least one test which selects the best recommendation out of several options (probably done by lowering the recommendation limit to like 2, and verify that some were selected over others)

707090 (changes requested):

